PR #2, opened 2024-06-22
Add books

StevenClontz:


StevenClontz:
@siwelwerd FYI - while I have all three books here, the CI/CD is about four minutes. The trick? I'm not generating images by default. (Moving checkit here is another todo.)

Everything will get built and deployed upon merge to main. (I'm guessing this will take 1-2 hours.) You can also manually request a full build on the PR, but you still have to wait for everything.

I'm going to continue working on this before we hit the edit-a-thon...

siwelwerd:
>FYI - while I have all three books here, the CI/CD is about four minutes

What is CI/CD?

StevenClontz:
https://en.m.wikipedia.org/wiki/CI/CD

Basically GitHub Actions magic that runs whenever we push stuff so we don't
have to do things manually.

PR #7, opened 2024-07-11
add checkit

StevenClontz:


StevenClontz:
Running a manual workflow at https://github.com/TeamBasedInquiryLearning/library/actions/runs/9901354483 (which should confirm the full builds and bank builds are working as intended).

StevenClontz:
> Running a manual workflow at https://github.com/TeamBasedInquiryLearning/library/actions/runs/9901354483 (which should confirm the full builds and bank builds are working as intended).

this works but the cloudflare preview isn't happy because checkit seed files are too big :-(

So I will need to disable cloudflare deployments for builds with checkit...

StevenClontz:
Attempting another manual workflow at https://github.com/TeamBasedInquiryLearning/library/actions/runs/10046975232 which should hopefully build the checkit (ensuring it builds, and can be deployed to GitHub Pages), but delete it before deploying to Cloudflare to avoid Cloudfile filesize limitation.

PR #61, opened 2024-07-22
moved last activity of EV4 to first of EV5

StevenClontz:
additional context/details...

### DIFF
--- a/linear-algebra/source/02-EV/05.ptx
+++ b/linear-algebra/source/02-EV/05.ptx
@@ -48,6 +48,27 @@
         </subsection>
 
 <subsection><title>Class Activities</title>
+
+    <activity estimated-time="5">
+        <statement>
+            <p>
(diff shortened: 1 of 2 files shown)


siwelwerd (commented):
I actually think this works better in the end of section EV04.

linear-algebra/source/02-EV/05.ptx
```diff
@@ -48,6 +48,27 @@
...
+        <li>
+            <p>
+                Yes
+            </p>
+        </li>
+        <li>
+            <p>
+                No
+            </p>
+        </li>
```
StevenClontz, line 67:
```suggestion
        <li>
            <p>
                Yeppers
            </p>
        </li>
        <li>
            <p>
                Nah man
            </p>
        </li>
```

StevenClontz: What if I do this @siwelwerd?

StevenClontz:
actually this is garbage

PR #72, opened 2024-07-22
Githook to Prevent/discourage committing to main directly

### LINKED ISSUE #70: Prevent/discourage committing to `main` directly
Reminded by #68 - we have this for `precalculus` so we should set it up here too.

StevenClontz:
Closes #70


StevenClontz:
FYI @siwelwerd

siwelwerd:
I tried this and now I can't push to any branch.  Getting this error

```
> git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file -
fatal: cannot run .githooks/pre-commit: No such file or directory
```

StevenClontz:
hrm. does `.githooks/pre-commit` exist?

siwelwerd:
Yeah, it apparently was having issues because there were CRLFs in there at the end of each line.  So it was trying to run it via /bin/bash^M which didn't exist.  See #73 for a fix, but please test that it doesn't break something else.

PR #73, opened 2024-07-22
Remove problematic CRLFs from pre-commit

siwelwerd:
This was breaking the githook on my Mac.  Please test that it doesn't break something else, all I did was remove the CRLFs at the end of each line.


StevenClontz (approved):
100% my fault, there's a subtle bug with `github.dev` (not Codespaces) that bit me here. I fixed it in another project at https://github.com/pi-base/data/blob/main/.vscode/settings.json so I should fix it here too.

PR #76, opened 2024-07-23
Tweaks to Precalc PR1 Checkit

siwelwerd:
Tweaks to fix https://github.com/TeamBasedInquiryLearning/precalculus/issues/359

### DIFF
--- a/precalculus/exercises/outcomes/PR/PR1/generator.sage
+++ b/precalculus/exercises/outcomes/PR/PR1/generator.sage
@@ -64,7 +64,9 @@ class Generator(BaseGenerator):
             face = "downwards"
             extrema = "maximum"
         
-        
+        form = choice(["standard","vertex"])
+        if form=="vertex":
+            q(x)=a*(x-h)^2+k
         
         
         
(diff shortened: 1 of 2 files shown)


precalculus/exercises/outcomes/PR/PR1/generator.sage
```diff
@@ -64,7 +64,9 @@ class Generator(BaseGenerator):
             face = "downwards"
             extrema = "maximum"
         
-        
+        form = choice(["standard","vertex"])
+        if form=="vertex":
```
StevenClontz, line 68:
```suggestion
        if choice(True,False):  # use vertex form instead
```

StevenClontz: take or leave this suggestion

StevenClontz (approved):

precalculus/exercises/outcomes/PR/PR1/generator.sage
siwelwerd: A bit late, but I actually think my code is more descriptive and like it a bit better.

Issue #51, opened 2024-04-02
Activity 3.8.1


AbbyANoble:
[Activity 3.8.1: The box](https://teambasedinquirylearning.github.io/calculus/2024e/AD8.html#activity-ad8-box) feels out of place here without some multiple choice to help guide students along. I'd recommending adding more to it, like [Activity 3.8.3](https://teambasedinquirylearning.github.io/calculus/2024e/AD8.html#activity-ad8-1) or putting it after [Activity 3.8.3](https://teambasedinquirylearning.github.io/calculus/2024e/AD8.html#activity-ad8-1).

PR #85, opened 2024-07-23
Add multiple choice and diagram to "The Box"

daverosoff:
Addresses https://github.com/TeamBasedInquiryLearning/library/issues/51.


StevenClontz:
We'll need a non-SVG version for print. The least annoying solution is to make a PNG that can be used in both HTML and PDF formats (but keep the source SVG here for the future).

calculus/source/03-AD/08.ptx
```diff
@@ -12,30 +12,96 @@
...
-            <p>Draw a diagram illustrating how the box is created.</p>
+            <statement>
+              <p>Draw a diagram illustrating how the box is created. Label the lengths of segments in the diagram and write <m>x</m> for the
+              sides of the smaller squares to be removed from the 8 by 8 square.</p>
+            </statement>
+          <solution>
+            <image source="thebox-1">
+              <shortdescription>Square of side length 8 with square corners of side length <m>x</m> removed</shortdescription>
+            </image>
+          </solution>
```
StevenClontz, line 36: For the purpose of a 4S in-class activity, I think providing four choices of diagrams and letting students choose which is most appropriate is better scaffolded, especially this early in a section.

daverosoff:
More like this? I think the pictures are OK, not sure the best way to give them labels A through D.

StevenClontz:
> More like this? I think the pictures are OK, not sure the best way to give them labels A through D.

Looks good - I'd either manually insert the letters in the images, or use a `<ol marker="A." cols="2"/>`.

daverosoff:
This still needs the PNG images which I have to make on the computer that has Inkscape.

daverosoff:
Closed per discussion at #124.

Issue #56, opened 2024-07-19
Finish Module ON


siwelwerd:
There is a partially written module on orthonormal bases, currently in `source/future-ON/`.  It would be great to complete this module for those that want to use this.

fragandi:
@jbunn3 This might be of interest

fragandi:
How is this going? @jbunn3 Would you like to connect today to talk about it?

jbunn3:
@fragandi Hi. It's starting well so far. Sure, I'm going to work a bit more on the first section this morning, but if you are interested in working on it too, we can talk today. I think Manoj might join in some tomorrow, but he has been busy teaching Precalc this week.

fragandi:
Great, let’s check in then later in the day… see you on the main call at 15:00 central = 3 pm Chicago time?

jbunn3:
Sounds good. I think Steven said 2:30 Chicago time. By the way, I created a Discussion #94 for this topic.

siwelwerd:
See https://github.com/orgs/TeamBasedInquiryLearning/projects/4 for progress on this.

jkostiuk:
Recording this publicly that I/we will prioritize this development over the Spring term. I invite folks interested in collaborating to reach out, in whatever capacity they can offer. 

A though that I've been wrestling with is this following: I personally incorporate orthogonality by including a sequence of problems in problem sets that develop properties of dot products, orthogonal subspaces, orthogonal bases, projections, etc. I've enjoyed introducing a little bit each time students learn a new thing. 

So, as we review, revise, and develop our working learning outcomes, I'm brainstorming the idea of a short 2-3 outcome module that could, potentially, be taught anytime after module EV (it could include dot-product/angle basics, orthogonal projections, orthogonal complements), and then perhaps we have some meaty applications that are more do-as-time-permits/instructor desires (that include say, Gram-Schmidt, orthogonal transformations). 

@siwelwerd @fragandi @jbunn3 @StevenClontz

PR #88, opened 2024-07-23
Fixed list style in Activity 6.1.1, added a space (typo), added Defn 6.1.4, added Activity after, modified text after

jbunn3:
This was my first attempt at making changes and creating new content. I will create this one as a draft to start.

### DIFF
--- a/linear-algebra/source/future-ON/01.ptx
+++ b/linear-algebra/source/future-ON/01.ptx
@@ -123,14 +123,14 @@ The dot product combines two vectors and creates a scalar that gives us geometri
 <activity estimated-time="5">
     <introduction>
         <p>
-            Consider the vectors <m>\vec u=\left[\begin{array}{c} 2 \\ 1  \end{array}\right]</m> and <m>\vec u=\left[\begin{array}{c} 1 \\ 3  \end{array}\right].</m>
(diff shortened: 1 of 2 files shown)


jbunn3:
Created a new commit to fix a couple mistakes that I had. This is beginning work for #56.

linear-algebra/source/future-ON/01.ptx
```diff
@@ -111,7 +111,42 @@ The dot product combines two vectors and creates a scalar that gives us geometri
...
+            </me>
+        </p></statement></definition>
+
+
+<activity estimated-time="5">
+    <introduction>
+        <p>
+            Consider the vectors <m>\vec u=\left[\begin{array}{c} 2 \\ 1  \end{array}\right]</m> and <m>\vec v=\left[\begin{array}{c} 1 \\ 3  \end{array}\right].</m>
+        </p>
+    </introduction>
```
StevenClontz, line 128: This activity (and the ones you didn't write before it) could use a bit more 4S scaffolding. My gut instinct is that at this stage, task A should be replaced with a given illustration to work off of. I think practicing drawing these sorts of pictures would work better as fluency builders or flexible extension.

StevenClontz (approved):
Thanks for this contribution! I'm approving as-is, but note my comment above.

In an ideal world, we are working from backwards design. So, something that would make reviewing these outcomes clearer would be to have example problems (if not a CheckIt outcome) that would be used to assess each learning outcome.

PR #89, opened 2024-07-23
add doenet interactives to GT1

StevenClontz:


siwelwerd (approved):
The bounding boxes on the interactives seem just a hair to small, so there are scroll bars that probably don't need to be there on most screens.  Going ahead and approving this, but if you have a way to fix that great, but if not just go ahead and merge.

StevenClontz:
> The bounding boxes on the interactives seem just a hair to small, so there are scroll bars that probably don't need to be there on most screens. Going ahead and approving this, but if you have a way to fix that great, but if not just go ahead and merge.

Limitation of PreTeXt, so merging as-is.

siwelwerd:
>Limitation of PreTeXt, so merging as-is.

I suspected as much.

PR #95, opened 2024-07-24
EQ4 template

AbbyANoble:
Template created for EQ4 CheckIt.


StevenClontz:
can you link to the discussion for this so I know how involved you want these polynomials ot be?

StevenClontz:
image

siwelwerd:
@StevenClontz I think you have mixed up PR4 and EQ4.  Improvements to PR4 belong elsewhere.

siwelwerd:
Relevant issue: https://github.com/TeamBasedInquiryLearning/precalculus/issues/394

siwelwerd:
> 🤦‍♂️

Example number 23,581 of why we review each others' work, even when we think we know what we are doing! :)

siwelwerd:
Re: The library of helpful functions

Should this live in `exercises/`, or more generally?  Because maybe this is helpful for https://github.com/TeamBasedInquiryLearning/precalculus/issues/424 as well.

StevenClontz:
image

AbbyANoble:
This is where I realize that the instructions should have been different for each one. 

Task 1: Solve for x.

Answer can just be like  x=1, 2  

Task 2: Keep as is!

StevenClontz:
> Re: The library of helpful functions
> 
> Should this live in `exercises/`, or more generally? Because maybe this is helpful for [TeamBasedInquiryLearning/precalculus#424](https://github.com/TeamBasedInquiryLearning/precalculus/issues/424) as well.

Probably more generally. I can take a swing at such a refactor.

StevenClontz:
> This is where I realize that the instructions should have been different for each one.
> 
> Task 1: Solve for x.
> 
> Answer can just be like x=1, 2
> 
> Task 2: Keep as is!

Looking back at the original issue, I think maybe we really want two parts for Task 2? That way we check if students can handle both inside and outside situations, and also strict and not-strict inequalities? I'll swing at that real quick now.

StevenClontz:
image

image

StevenClontz:
Merged so I can work on tidying up the new Sage library

PR #96, opened 2024-07-24
Cg2wilson-DF-typos

cg2wilson:
Changed the \frac to \dfrac in most instances in chapter DF. \frac not changed in exponents of functions, or in complex fractions (e.g. Activity 2.2.10)

### DIFF
--- a/calculus/source/02-DF/01.ptx
+++ b/calculus/source/02-DF/01.ptx
@@ -91,7 +91,7 @@
         <me> \frac{f(2+h)-f(2)}{(2+h)-2} = \frac{f(2+h)-f(2)}{h}.</me>
     </observation>
     
-    <observation><p>We want to be able to consider intervals before and after <m>t=2</m>. A positive value of <m>h</m> will give an interval after <m>t=2</m>. For example, the interval <m>[2,3]</m> for <m>h=1</m>. A negative value of <m>h</m> will give an interval before <m>t=2</m>. For example, the interval <m>[1,2]</m> corresponds <m>h=-1</m>. In the formula above, it looks like the interval would be <m>[2,1]</m>, but the standard notation in an interval is to write the smallest number first. This does not change the difference quotient because </p>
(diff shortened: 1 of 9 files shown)


calculus/source/02-DF/01.ptx
```diff
@@ -112,17 +112,17 @@
     </activity>
     
  <definition xml:id = "def-inst-vel"> 
-     <p> The instanteous velocity at <m>t=a</m> is the limit as <m>h \to 0</m> of the difference quotient <m>\frac{f(a+h)-f(a)}{h}</m>. In the activity above the instantaneous velocity at <m>t=2</m> is given by the limit
+     <p> The instanteous velocity at <m>t=a</m> is the limit as <m>h \to 0</m> of the difference quotient <m>\dfrac{f(a+h)-f(a)}{h}</m>. In the activity above the instantaneous velocity at <m>t=2</m> is given by the limit
```
StevenClontz, line 115:
```suggestion
     <p> The <term>instantaneous velocity</term> at <m>t=a</m> is the limit as <m>h \to 0</m> of the difference quotient <m>\dfrac{f(a+h)-f(a)}{h}</m>. In the activity above the instantaneous velocity at <m>t=2</m> is given by the limit
```

StevenClontz: You didn't do this, but I'd prefer the "activity above" comment to be separated out into a separate observation/remark.

Another enhancement while we're here: use $\lim_{h\to 0}$ notation in the definition, not just words.

calculus/source/02-DF/01.ptx
```diff
@@ -425,7 +425,7 @@ Defined and continuous on the interval <m>[-5,5]</m>.
                         </li>
                         <li>
                             <p>
-<m>\displaystyle\lim_{h\to 0} \frac{f( 2 +h)-f( 2 )}{h}&lt; 0</m>
+<m>\displaystyle\lim_{h\to 0} \dfrac{f( 2 +h)-f( 2 )}{h}&lt; 0</m>
```
StevenClontz, line 428: Another potential enhancement, rather than littering `\displaystyle`s for limits, we could try something like https://tex.stackexchange.com/questions/161073/always-using-displaystyle-only-for-lim

StevenClontz (approved):
Thanks! I left some comments for potential future work.

PR #109, opened 2024-07-25
First stab at features subsection in linear algebra book

fragandi:
We already had an instructor section and a videos section. I think these could all be merged into one, but for now I just added a new Features of the text section in the front matter to start brainstorming this


siwelwerd:
I like what you've written.  I agree, lets take all those sections and merge them into one, but with a subsection for each of those.

siwelwerd:
As part of this, lets update the link to the TBIL Resource library.

siwelwerd (changes requested):
We should combine the different pages into one page, and pull it out of the LA book to the top level so the same info can be included into all books.

siwelwerd:
Hmm, I had an older version of PreTeXt and it worked fine, 2.6.0 is throwing
```
error: * PTX:BUG:  development bug, new level does not match old level for "section"
*              located within: "preface-community" (xml:id), "Community and Support" (title)
```

It seems to object to a section within a preface?

StevenClontz:
Apparently not: https://pretextbook.org/doc/schema/

`paragraphs` with a title should work

StevenClontz (approved):

Issue #49, opened 2024-03-04
AD3 Activity


AbbyANoble:
In [Activity 3.3.11(c)](https://teambasedinquirylearning.github.io/calculus/2024e/AD3.html#right-triangles-related-rates-4), the way it's worded makes it sound like we are looking for **the** maximum rate of change when I think the intention is to ask which of the choices has the largest rate of change. 

In addition, I would recommend changing the wording at the beginning of the activity about the Pythagorean theorem. It seems weird to say _also_ known. It's primarily known as that in the western world. And while I appreciate pointing out that Pythagoras is not solely responsible for that theorem, it feels out of place where it is. 

I would recommend a remark similar to [Remark 3.3.8: Volume formulas](https://teambasedinquirylearning.github.io/calculus/2024e/AD3.html#rmk-volume-formulas) describing the theorem. (Or a theorem environment stating the theorem.) With a remark or "fun fact" after about how the result was well known before him.

fragandi:
As the original author, I agree with your points and will take a stab at editing this!

fragandi:
I did not want to bring more attention to the side remark with a new environment, but I have added a separate remark for the theorem and rephrased the side remark with more of a curiosity tone. 

I have also rewritten part c), let me know if this is good enough!

jbunn3:
I recommend that we keep the file as is for now for the 2025 edition, since we have the answers in there now. But I think we can incorporate these changes for the next edition. I think we'll need to create a new PR or merge the branch from #110 with main before trying to continue.

PR #110, opened 2024-07-25
First attempt at addressing AD3 Activity #49

fragandi:
Broke theorem from activity and reworded part c) to emphasize we are asking when it is largest (even though a completely legit thing is to compute the rate in each case to figure out which one is largest)

### DIFF
--- a/calculus/source/03-AD/03.ptx
+++ b/calculus/source/03-AD/03.ptx
@@ -279,40 +279,41 @@ recall that when <m>y</m> is a function of <m>x</m>, which in turn is a function
     
 <!-- Using triangles     -->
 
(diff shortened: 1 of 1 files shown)


cg2wilson (commented):
What's the purpose for killing the `<statement>` tags in the `<task>`s?

fragandi:
Not sure I know, probably just bad syntax from my part. I was more worried about whether the change addressed fully the original issue.

siwelwerd:
Try to include the `<statement>`, as if an `<answer>` is added later this will cause an issue with the lack of a `<statement>`.

StevenClontz:
Is this still viable? Looks like the file was updated by a separate PR.

PR #131, opened 2024-07-26
Added a fact to be used later for computing orthogonal projections.

jbunn3:
We will need a definition of transpose to be added later too. I'll make an issue for that.

### DIFF
--- a/linear-algebra/source/future-ON/02.ptx
+++ b/linear-algebra/source/future-ON/02.ptx
@@ -51,9 +51,9 @@
         <fact>
             <statement>
                 <p>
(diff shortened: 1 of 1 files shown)


jbunn3:
The goal here was to get a little more practice in GitHub and Codespaces by adding in some content. I don't think ON2 (or ON3) will be in a usable form for a while. But hopefully, we can get ON1 ready soon as a sample section to test out this fall for someone.

StevenClontz (approved):

StevenClontz:
compare with [Definition 1.1.1](https://tbil.org/linear-algebra/preview/instructor/LE1.html#LE1-3-2) 
 where we kind-of define transpose for a euclidean vector

PR #135, opened 2024-07-26
We are planning to add new Sec on Linear Algebra.

mlamich1:


jbunn3:
@mlamich1 It looks like you just added in that example had. We can certainly consider adding this in as another fluency builder. Were you also going to do a pull request for the law of cosines content?

mlamich1:
Yes, I have almost done with Law of Cosine and drafted it. Will be pushing shortly.

mlamich1:
@jbunn3 I just added angle between two vectors as draft. Can you see from your end?

jbunn3:
@mlamich1 Yeah, I see it on the other pull request.

PR #140, opened 2024-07-26
LF2 CheckIt template

AbbyANoble:
Closes https://github.com/TeamBasedInquiryLearning/library/issues/173


siwelwerd:
The relevant changes are all in the most recent commit c33e392 . If this is too messy we can just make a clean PR.  The Internet led me to believe that I did the "right" thing to this, but I am not convinced that was correct.

siwelwerd:
Yeah, sorry.  It says it should merge fine, but a clean PR is probably easier.  I did that for the next one BTW.

image

PR #143, opened 2024-07-26
EL2 CheckIt template

AbbyANoble:
see https://github.com/TeamBasedInquiryLearning/library/issues/182

Also there were some errant spaces in the bank.xml that I removed.


AbbyANoble:
Can ignore the bank.xml changes in my commit as they were merged with #146

PR #147, opened 2024-07-26
PR5 Checkit updates

siwelwerd:
Closes https://github.com/TeamBasedInquiryLearning/library/issues/174.  I ended up refactoring the whole thing, potentially this should get pulled into the common sage file.

image


precalculus/exercises/outcomes/PR/PR5/template.xml
```diff
@@ -1,66 +1,23 @@
...
-                source="assets/PR4/generated/{{__seed__}}/plot.png"
-                description="The plot of y=h(x), along with special points."/>
-            </p>
-        </content>
-        <outtro>
-            <p>
-                <m>h(x)={{h}}</m>
-            </p>
+        <p>
+            The {{name}} of <m>{{fname}}(x)={{f}}</m> are <m>x  = <!--{{#roots}}--> {{r}},  <!--{{/roots}}--> </m>.
```
siwelwerd, line 17: @StevenClontz Is there a clever way to get the list `{{r}}` separated by commas but not have a comma at the end?  The only idea I have is to do it on the python side, and then just pass a string here instead of a list.

StevenClontz: Either that, or I've used `<list>{{#things}}<item>{{thing}}</item>{{/things}}</list>` if a bulleted list works

StevenClontz: yeah here I'd use `",".join(roots)` on the Python side (or whatever)

siwelwerd: Went with `",".join([latex(r) for r in roots])`. Should be good to go now.

StevenClontz:
Is this really what is asked of college algebra / precalculus students? @kathypinzon 

Are we sure we don't want something more scaffolded? This seems way more intense than [Activity 4.5.18](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/PR5.html#poly-degree-4).

siwelwerd:
>Is this really what is asked of college algebra / precalculus students?

Maybe y'all can discuss tomorrow.  But trying to mix in non-integer rationals + either irrational or complex roots as indicated in https://github.com/TeamBasedInquiryLearning/precalculus/issues/413 will necessarily yield these gnarly-looking polynomials.

StevenClontz:
We discussed in our meeting some extra constraints for this to get the numbers to be a little less foreboding. I'll work on this now.

StevenClontz:
I think this looks like what we discussed this afternoon. @kathypinzon @tdegeorge ?

image

StevenClontz:
Using plots to help find rational roots (but ignoring the rational root theorem, too many choices!), I was able to factor these with the minimum pain required to perform long polynomial division.

If you want to assess understanding of the rational root theorem, maybe add a Task 3: "Use the rational root theorem to explain why N might be a root of F(X), but M cannot be."?

StevenClontz (approved):

StevenClontz:
Going ahead and approving (we need a fix I made to `preview_exercises.py`) but open a new issue if we want to make further changes.

PR #151, opened 2024-07-26
Changed section titles to title case

cg2wilson:


AbbyANoble:
We should probably update the Check It bank as well to match.

siwelwerd (approved):

PR #153, opened 2024-07-26
Added an observation to ON 2

jbunn3:
I also added a commented out placeholder for an activity idea before the observation.

### DIFF
--- a/linear-algebra/source/future-ON/02.ptx
+++ b/linear-algebra/source/future-ON/02.ptx
@@ -63,6 +63,19 @@
             </statement>
         </fact>
 
+        <!-- <p>
+            Insert scaffolded activity that has them compute both sides of the equation from the above fact for a given standard matrix <m>A</m> with one column and a given vector <m>\vec u</m>. Finally,
+            they will solve for <m>\vec x</m> to see the following observation.
+        </p> -->
+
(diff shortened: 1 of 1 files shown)


linear-algebra/source/future-ON/02.ptx
```diff
@@ -63,6 +63,19 @@
...
+            they will solve for <m>\vec x</m> to see the following observation.
+        </p> -->
+
+        <observation>
+        When the image of a linear transformation is one-dimensional, then its standard matrix <m>A</m> only has one column <m>\vec v</m>,
+        and for any vector <m>\vec u</m>,
+        <me>
+            \vec u_W=\dfrac{\vec u\cdot \vec v}{\vec v\cdot \vec v}\vec v.
+        </me>
+        </observation>
```
StevenClontz, line 77:
```suggestion
        <observation>
            <p>
        When the image of a linear transformation is one-dimensional, then its standard matrix <m>A</m> only has one column <m>\vec v</m>,
        and for any vector <m>\vec u</m>,
        <me>
            \vec u_W=\dfrac{\vec u\cdot \vec v}{\vec v\cdot \vec v}\vec v.
        </me>
            </p>
        </observation>
```

Issue #154, opened 2024-07-26
Add LE2 warmup/cooldown and define coefficient matrix


StevenClontz:
See also https://github.com/TeamBasedInquiryLearning/linear-algebra/pull/382/files - we should also define the coefficient matrix.

StevenClontz:
Closed by #156

PR #156, opened 2024-07-26
LE2 warmup

StevenClontz:
- reimplement https://github.com/TeamBasedInquiryLearning/library/issues/154
- define coefficient matrix

### DIFF
--- a/linear-algebra/source/01-LE/01.ptx
+++ b/linear-algebra/source/01-LE/01.ptx
@@ -630,9 +630,8 @@ Augmented matrix:
     <definition xml:id="LE1-def-coefficient-matrix">
         <statement>
             <p>
-We won't always care to know the value of the constant terms <m>b_i</m>
-(for example, when they are all <m>0</m>).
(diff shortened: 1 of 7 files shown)


linear-algebra/source/01-LE/01.ptx
```diff
@@ -561,6 +626,38 @@ Augmented matrix:
         </p>
         </statement>
     </example>
+
+    <definition xml:id="LE1-def-coefficient-matrix">
+        <statement>
+            <p>
+We won't always care to know the value of the constant terms <m>b_i</m>
+(for example, when they are all <m>0</m>).
+In this case, we may use a <term>coefficient matrix</term>:
```
StevenClontz, line 635:
```suggestion
In the case we need to only analyze the coefficients of a linear system,
we will use a <term>coefficient matrix</term> rather than the full augmented matrix:
```

jkostiuk (approved):
approved live with steven

PR #157, opened 2024-07-26
FN5 CheckIt template

AbbyANoble:
see https://github.com/TeamBasedInquiryLearning/library/issues/168


precalculus/exercises/outcomes/FN/FN5/template.xml
```diff
@@ -1,24 +1,112 @@
...
+                    <p>
+                        <m>{{difference}}</m>
+                    </p>
+                </outtro>
+                
+            </knowl>
+            <knowl>
+                <content>
+                    <p>
+                        <m>{{fname1}}(x) \times {{fname2}}(x)</m>
```
StevenClontz, line 41:
```suggestion
                        <m>{{fname1}}(x) \cdot {{fname2}}(x)</m>
```

Maybe - @AbbyANoble which notation is standard in this course?

precalculus/exercises/outcomes/FN/FN5/generator.sage
```diff
@@ -1,10 +1,40 @@
...
+        "funct2": f2,
+        "funct3": f3,
+        "funct4": f4,
+        "fname1": fname1,
+        "fname2": fname2,
+        "fname3": fname3,
+        "fname4": fname4,
+        "sum": f1+f2,
+        "difference": f1-f2,
+        "product": f1*f2,
```
StevenClontz, line 34:
```suggestion
        "product": expand(f1*f2),
```

Again maybe - the sum and difference are simplified so it's not clear to me the point here.

Of course we could make the denominator function always be monomial, so the quotient can always be simplified (modulo zero-denominator situations I guess).

StevenClontz (approved):

precalculus/exercises/outcomes/FN/FN5/generator.sage
siwelwerd: Specific request in #168:
>Answers not completely multiplied out and simplified if possible. Have them still look similar to the outside functions

StevenClontz (approved):
Sounds good. @AbbyANoble let us know if you prefer $\cdot$ over $\times$ for multiplication, but going ahead and merging.

PR #158, opened 2024-07-26
FN2 Checkit Precalculus

### LINKED ISSUE #170: FN2 CheckIt
Reviewer: @kathypinzon 

**Task 1**  modeled after [Activity 2.2.7](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/FN2.html#activity-fn-values).  Except just give one function (randomly chosen from a polynomial, piecewise, radical, or rational) and ask for f of two values. Occasionally throw in something undefined. 

**Task 2** modeled after [Activity 2.2.13](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/FN2.html#activity-fn-graphical). Give a graph (polynomial, piecewise, radical, or rational, reuse randomization from Task 1 maybe?), then ask: 

- Find f(some value).
- For which x-value(s) in the interval [ __ , __ ] does f(x)= some value?  

The second question should have between 0 and 3 answers, changing depending on the problem. Interval can depend on the function, or fix to something like [-10,10]. Can ask students to estimate if needed like part (d) of the activity if it's too contrived to fix everything to an integer or integer.5 value that's easy to see in a graph.

siwelwerd:
Closes #170

### DIFF
--- a/precalculus/exercises/outcomes/FN/FN2/generator.sage
+++ b/precalculus/exercises/outcomes/FN/FN2/generator.sage
@@ -46,38 +46,17 @@ class Generator(BaseGenerator):
 
 
     R.<z> = PolynomialRing(QQ)
-    count = choice([0..3])
-    xpoints = sample([-8..8],3)
+
+    level=choice([-10..-1,1..10])
+    xpoints=sample([-8..8],3)
     xpoints.sort()
-    ypoints = sample([-20..20],2)
-    ypoints.sort()
-    ypoints.append(choice([-20..(ypoints[1]-1)]))
+    points = [ (p,level) for p in xpoints]+[(xpoints[0]-choice([2..4]), level+choice([-1,1])*choice([3..5])) ]
+    g=R.lagrange_polynomial(points)
 
-    level=choice([-10..-1,1..7])
-    ypoints=sample([(level+1)..12],3)
-    ypoints.sort()
-    xpoints=[]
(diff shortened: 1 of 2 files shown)


siwelwerd:
@StevenClontz @AbbyANoble Could use help coming up with a clever way to generate more graphs meeting [your critera](https://github.com/TeamBasedInquiryLearning/precalculus/issues/326) for Task 2.

siwelwerd:
>@StevenClontz @AbbyANoble Could use help coming up with a clever way to generate more graphs meeting https://github.com/TeamBasedInquiryLearning/precalculus/issues/326 for Task 2.

Idea stolen from @matthematician's college algebra bank: Use [Lagrange interpolation](https://doc.sagemath.org/html/en/reference/polynomial_rings/sage/rings/polynomial/polynomial_ring.html#sage.rings.polynomial.polynomial_ring.PolynomialRing_field.lagrange_polynomial)

siwelwerd:
Need to figure out how to add some gridlines, and add more randomization.  Right now it always has 3 points in Task 2.2.  Also having trouble making Task 2.1 provide an integer solution without breaking things.

siwelwerd:
Completely rewrote what I was doing in Task 2.  It uses lines instead of curves, but that seems okay for this learning outcome. Occasionally (but not usually), the students are asked to find a point that is not an integer, so it rounds answers to a single decimal place.

PR #195, opened 2024-07-30
Common colophon, update precal content

StevenClontz:


StevenClontz:
whoops meant to do the precal stuff in a separate PR but going ahead here anyway

PR #202, opened 2024-07-30
fix build_bank.py

StevenClontz:


StevenClontz:
Testing at https://github.com/TeamBasedInquiryLearning/library/actions/runs/10171233278

PR #207, opened 2024-07-31
clarify guess-check vs methodical methods in EV2

### LINKED ISSUE #27: Suggestion for Activity 2.2.5
Activity 2.2.5 is a scaffolded discussion guiding students towards the presence of a row of 0s determining if a set of vector spans the whole vector space. As written, parts (a) and (b) are a good warmup to have students decide if 2 given vectors are in the span or not. However, I'm wondering if changing (b) so that the answer to both (a) and (b) is "yes" might be more helpful? 

As it is written now, the answer to the main question "do these vectors span R^3" is "no" because of (b)--parts (c) and (d) are not needed. 

If the answer to (b) was also "yes", then when we start part (c), we're still in the "well, we can't fully conclude one way or the other yet, so let's keep going"...

StevenClontz:
Closes #27

### DIFF
--- a/source/linear-algebra/source/02-EV/02.ptx
+++ b/source/linear-algebra/source/02-EV/02.ptx
@@ -196,6 +196,11 @@ since <m>\IR^1=\setBuilder{cx}{c\in\IR}</m>.
                 \left[\begin{array}{c}-2\\0\\1\end{array}\right],\left[\begin{array}{c}-2\\-2\\2\end{array}\right]\right\}</m>. 
             </p>
         </statement>
+        <answer>
+            <p>
+The vector belongs to the span.
+            </p>
+        </answer>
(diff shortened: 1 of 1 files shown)


StevenClontz:
@jkostiuk take a look at https://20bebcb4.tbil.pages.dev/linear-algebra/preview/instructor/EV2#EV2-activity-guess-check and the following activity - what do you think?

jkostiuk:
I like it! Cheers,JordanOn Jul 31, 2024, at 4:04 PM, Steven Clontz ***@***.***> wrote:﻿
@jkostiuk take a look at https://20bebcb4.tbil.pages.dev/linear-algebra/preview/instructor/EV2#EV2-activity-guess-check and the following activity - what do you think?

PR #214, opened 2024-08-02
Precalculus EQ5 Checkit

### LINKED ISSUE #188: EQ5 Check It
https://github.com/TeamBasedInquiryLearning/precalculus/discussions/339

Task 1: There are many ways to solve a quadratic equation, but sometimes it is better to use one particular method. For each equation, determine which method would be the best to use and why. You must choose a different method for each equation and cannot use one method more than once. (All equations having rational or integer solutions?)

quadratic equation with a=1 (factorable). Example: x^2+4x=-3
quadratic equation with a binomial (square root property): Example: 10(t+6)^2+36=90
quadratic equation with a=1 (complete the square - middle term can easily take half of and square): Example: a^2+14a=51
quadratic equation with a not equal to 1 and not factorable (quadratic formula): 3x^2-8x=-4
Task 2: Choose two of the equations above and solve using the method you stated is the best to use. Do you still think that method was the best? Why or why not?

siwelwerd:
Closes #188

### DIFF
--- a/source/precalculus/exercises/outcomes/EQ/EQ5/generator.sage
+++ b/source/precalculus/exercises/outcomes/EQ/EQ5/generator.sage
@@ -1,10 +1,40 @@
+load("../sage/common.sage")
+
 class Generator(BaseGenerator):
-    def data(self):
-      return {
-        "eq": 3,
-        "ineq": 4,
-        "eqansinterval": 5,
-        "eqansgraph": 6,
-        "ineqansinterval": 7,
-        "ineqansgraph": 8
-      } 
+  def data(self):
+    
+
+    methods=["factoring", "the square root property", "completing the square", "the quadratic formula"]
+    equations =[]
+
+    #Factoring
+    r1=choice([-1,1])*choice([2..6])
+    r2=choice([-1,1])*choice([2..6])
+    equations.append( { "equation":  x^2-(r1+r2)*x==-1*r1*r2, "method":"factoring", "roots": f"{r1}\\text{{ and }}{r2}"})
+
+    #Square root property
+    a=choice([2..5])
+    h=choice([-7..-1,1..7])
+    c=choice([2..5])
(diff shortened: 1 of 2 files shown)


AbbyANoble:
Checking in to see if this will be merged to the bank soon! It's not there currently.

siwelwerd:
If you want to approve any of these, the easiest way to preview them is to hit `,` (comma) from this page, which will fire up a codespace on the appropriate branch. Then in the terminal run `python scripts/preview_outcome.py precalculus EQ5`, which will generate the problems and give you a link to a preview webpage.  Change EQ5 to the appropriate outcome.  Note that it can be slow to run the generation

I'm happy to merge these once someone else lays eyes on them, that doesn't necessarily have to be @StevenClontz (though he's welcome to review and merge these too).

AbbyANoble:
Ok. Cool. I didn't know how to preview these because the links didn't work for me.

Unfortunately, I am out of storage space for the month so I can't start a codespace to see them. 🤪

siwelwerd:
Yeah, Steven and I were talking the other day about a way to incorporate the previews here through Cloudflare.  I think our conclusion was it's doable, but Steven will need a bit of time to make it happen.

StevenClontz:
image

Why is the first best solved by factoring, and the second best solved by completing the square? I think the other two strategies are more clearly distinguished. I see this is similarly unclear in #188's notes...

Maybe factoring problems should always have $b$ odd in $ax^2+bx+c$ to make completing the square obnoxious.

StevenClontz (changes requested):
consider my recent comment

siwelwerd:
Yeah, I think in the context of the learning outcome it's not a huge deal as they are instructed to choose a different method for each of the four.  But it was easy enough to make sure the sum of the roots is always odd in the factoring one.

StevenClontz (approved):

PR #216, opened 2024-08-02
Precalculus EQ7 Checkit

### LINKED ISSUE #187: EQ7 CheckIt
Reviewer: @tdegeorge 

Instructions: Solve the following inequalities. Express your answers in interval notation and graphically on a number line.

Two tasks in the problem. One quadratic and one rational. In both tasks, vary use of <, >, <=, and >= (even though I don't in my examples).

**Task 1**  is a quadratic inequality that can be factored fairly easily. Pull from examples like the following (but without any scaffolding):

- [Activity 1.7.5](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/EQ7.html#quadratic-inequality-coeff-less-than-0)
- [Activity 1.7.6](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/EQ7.html#quadratic-inequality-coeff-greater-than-or-equal-0) 
- ~~[Activity 1.7.2](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/EQ7.html#quadratic-inequality-coeff-greater-than)~~ 

**Task 2**  is a rational inequality. Stick to whole numbers for the partition numbers. Pull from examples like the following:

- 12/(x-5) > 4
- (x+6)/(x+1) < 2
- (3x+4)/(x-4) > x-1
- x > 8/(2+x)
- 10/(x+4) > 3-x
- -1/(x-6) > 2/(9-x)

siwelwerd:
Closes #187

### DIFF
--- a/source/precalculus/exercises/outcomes/EQ/EQ7/generator.sage
+++ b/source/precalculus/exercises/outcomes/EQ/EQ7/generator.sage
@@ -79,10 +79,7 @@ class Generator(BaseGenerator):
 
     
     return {
-      #"eq": f"\dfrac{{{latex(a*x+b)}}}{{{latex(x+c)}}} {ineq} \dfrac{{{latex(d*x+e)}}}{{{latex(x+f)}}}",
       "eq": f"{latex(LHS)} {(ineq)} {latex(RHS)}",
-      "cut": partition_points,
-      "numberline":numberline,
       "interval_string": "\\cup".join(intervals),
       "intervals": intervals,
     } 
--- a/source/precalculus/exercises/outcomes/EQ/EQ7/generator.sage
+++ b/source/precalculus/exercises/outcomes/EQ/EQ7/generator.sage
@@ -1,10 +1,55 @@
+load("../sage/common.sage")
+
 class Generator(BaseGenerator):
-    def data(self):
-      return {
-        "eq": 3,
-        "ineq": 4,
-        "eqansinterval": 5,
-        "eqansgraph": 6,
-        "ineqansinterval": 7,
-        "ineqansgraph": 8
-      } 
+  def data(self):
+
+    partition_points=[]
+    equation_type=choice(["linear", "quadratic"])
+
+    if equation_type=="linear":
+      #Ensure a=d
+
+      a=choice([0..6])
+      d=a
+      c=choice([-6..-1,1..6])
+      e=choice([-6..-1,1..6])
+      f=choice([-6..-1,1..6])
+      b=(f-1)*a*(f-c)+e*(c-f+1)
+      partition_points = [1-f]
+
+    else:
+      #Quadratic
+      r1,r2 = sample([-5..-1,1..5],2)
+      b=1
+      e=choice([-6..6])
+      c=choice([-6..6])
+      f=r1*r2+e*c
+      a=(-r1-r2+e-c-b)/(f-c)
+      d=a-1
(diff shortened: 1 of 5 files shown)


siwelwerd:
Okay, finished Task 2.  Now to handle Task 1 🥵

siwelwerd:
This could probably be refactored to be nicer, but I am tired of it and there are plenty of other things to work on now.  It seems to do what we want it to do though, so marking this ready.

StevenClontz:
image

Something is goofy, I don't think 0 is a solution to either inequality?

StevenClontz (changes requested):
see recent comment

siwelwerd:
Might just be a backwards inequality, so hopefully an easy fix.

source/precalculus/sage/common.sage
```diff
@@ -72,4 +121,13 @@ class TBILPrecal:
...
                       imaginary_part=[i for i in range(-5,6) if i != 0],length=1):
         '''Generates a list of unique 2-tuples of conjugate pairs of complex (not real) numbers of the form a+bi.'''
-        return sample([(a+b*I,a-b*I) for a in real_part for b in imaginary_part],length)
\ No newline at end of file
+        return sample([(a+b*I,a-b*I) for a in real_part for b in imaginary_part],length)
+
+    @staticmethod
+    def check(LHS,operator, RHS):
+        if operator == '<': return LHS > RHS
+        if operator == '>': return LHS < RHS
```
siwelwerd, line 129: >
```suggestion
        if operator == '<': return LHS < RHS
        if operator == '>': return LHS > RHS
```

siwelwerd: Pretty sure this will fix the issue @StevenClontz spotted, but this should probably just be refactored along the lines of #257

siwelwerd:
I refactored this to use Sage inequalities.  I was basically using the same code twice to generate intervals for the solutions, so I factored that into a helper function in common.sage.  This makes it much clearer to read.   I also managed to catch a divide-by-zero error and fix that too.

source/precalculus/exercises/outcomes/EQ/EQ7/generator.sage
```diff
@@ -1,10 +1,81 @@
...
+    RHS = (expand((d*x+e)*a.denominator())).mul(1/(x+f),hold=True)
+    ineq=choice(["<","<=",">",">="])
+    if ineq == "<":
+      rational_inequality = LHS < RHS
+    elif ineq == "<=":
+      rational_inequality = LHS <= RHS
+    elif ineq == ">":
+      rational_inequality = LHS > RHS
+    elif ineq == ">=":
+      rational_inequality = LHS >= RHS
```
siwelwerd, line 59: I couldn't see a clean way to use `CheckIt.shuffled_inequality()` here, as all I have to play with are two rational functions, and I just want $P(x)<Q(x)$ or similar. The `shuffled_inequality()` kept doing silly things like $P(x)+Q(x)>0 $which I didn't want.

siwelwerd: After stewing on this over the weekend, thought up a cleaner fix implemented in [ac06620](https://github.com/TeamBasedInquiryLearning/library/pull/216/commits/ac0662031f5bb64b7191cbbb589ee77c321cf1dd) using the `operator` module.

StevenClontz (approved):

Issue #228, opened 2024-08-13
EQ1 notation issue


AbbyANoble:
The linear inequalities are showing up as $<=$ and $>=$ instead of $\leq$ and $\geq$.

siwelwerd:
Reasonably confident this is in multiple outcomes, so edited the title to reflect this.  Probably need to just write a quick function that converts `"<="` to `r"\leq"`

siwelwerd:
#229 takes care of this in EQ1, but I suggest leaving this open until all of the outstanding Checkit PRs are merged, and then audited for this same issue.

siwelwerd:
EQ7 needs to be updated too, but I can't figure out how to cleanly do that since #216 is pending.  Should be a fix quick once both #229 and #216 are merged.

StevenClontz:
Inequalities are objects Sage understands, so `return {"ineq": x < 0}` should Just Work. I'll look into this more close when I have a chance though.

Note also the `CheckIt.shuffled_inequality` function.

siwelwerd:
It would be good to refactor to use that in the long run, but if the results are adequate as is this will fix the display with minimal additional effort.

siwelwerd:
This was dealt with by #257 and #216.  Reopen if you spot this anywhere else but I think we got them all.

PR #229, opened 2024-08-13
Convert <= to \leq, etc.

siwelwerd:
Addresses #228 .  Added a helper function and implemented it in EQ1.

### DIFF
--- a/source/precalculus/sage/common.sage
+++ b/source/precalculus/sage/common.sage
@@ -85,5 +85,4 @@ class TBILPrecal:
             return r"\lt"
         elif ineq==">":
             return r"\gt"
-        else:
-            return "Not an inequality symbol"
+        raise ValueError("Not an inequality symbol")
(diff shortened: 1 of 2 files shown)


source/precalculus/sage/common.sage
```diff
@@ -72,4 +72,18 @@ class TBILPrecal:
...
+        if ineq=="<=":
+            return r"\leq"
+        elif ineq==">=":
+            return r"\leq"
+        elif ineq=="<":
+            return r"\lt"
+        elif ineq==">":
+            return r"\gt"
+        else:
+            return "Not an inequality symbol"
```
StevenClontz, line 89:
```suggestion
        raise ValueError("Not an inequality symbol")
```

StevenClontz: We should also contrast with `CheckIt.shuffled_inequality` used at https://github.com/TeamBasedInquiryLearning/library/blob/fa1e46b57cde6a08ce5735a42366b9e40c4dd314/source/precalculus/exercises/outcomes/EQ/EQ4/generator.sage#L29 see also https://github.com/StevenClontz/checkit/blob/911954f9b98df090d2ddb4dbe4f0e2a97bd1c96d/dashboard/checkit/wrapper/wrapper.sage#L35

PR #233, opened 2024-08-15
update from library main

StevenClontz:


StevenClontz:
Assuming the yellow dot 🟡 above becomes a green checkmark ✅ and not a red X ❌ then this is ready to merge.

PreTeXt prefaces are limited and cannot have sections/subsections. I tried to fix the errors in syntax but the scaffolding you want isn't quite there. Maybe the syllbaus could be in the appendix (I think that allows for subsectioning).

StevenClontz:
Oof, this was meant for Jordan's branch...

StevenClontz:
Reimplemented for Jordan at https://github.com/jkostiuk/brown-tbil/pull/2

PR #238, opened 2024-08-20
Precalculus LF4 Checkit

### LINKED ISSUE #184: LF4 Check It
Reviewer: @AbbyANoble 

UPDATED 
Make sure that if one task asks for parallel, the other task asks for perpendicular (and vice versa).

Task 1: 
Find the value of a so that the line passing through the points (2, a) and (-3, 3) is perpendicular/parallel to the line that passes through the points (4, 6) and (-1,1).

Vary which one is being asked (parallel or perpendicular). Fractional slopes are okay too.

Task 2: 
Similar to [Activity 3.4.10](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/LF4.html#write-equations-for-both-parallel-and-perpendicular). Give an equation and a point and ask students to find an equation of the line that is either parallel or perpendicular to the given line and passes through a point.

siwelwerd:
Closes #184


siwelwerd:
Oops.  Branched this off of the LF3 branch instead of off main.  Should have just been the last commit. Let me see if I can fix that...

StevenClontz:
image

Double-check my work  - I'm getting a slope of $-7/8$ for the line defined by the points, so we want a perpendicular slope of $+8/7$, which I think yields $a=-40/7$ instead?

siwelwerd:
So it looks like this is the case `orientation1=="perpindicular"` and `solve_component=="x"`.  It looks like there is one other one of that case in the set, and it looks okay?

image

siwelwerd:
>Double-check my work

I did and thought you were right, but now I am convincing myself that this problem is actually okay.  Isn't the slope between (-35/8,-4) and (0,1)  $\frac{1--4}{0--\frac{35}{8}} = \frac{5}{\frac{35}{8}} = \frac{40}{35} = \frac{8}{7}$ ?

StevenClontz (approved):
I goofed in my solution, this looks good.

PR #160, opened 2024-07-29
Precaclulus FN3 Checkit

siwelwerd:
Eventually will close https://github.com/TeamBasedInquiryLearning/library/issues/169

PR #254, opened 2024-08-23
Re-implement #160

### LINKED ISSUE #169: FN3 Check It
Reviewer: @kathypinzon 

Task 1

https://matthematician.github.io/collegealgebra/#/bank/FN5/1/

Screenshot 2024-06-19 at 7 04 55 PM

Task 2

Give a polynomial graph (so as to have more than one max/min) and ask to find all characteristics (domain, range, x-intercepts, y-intercepts, relative max/min, global max/min, intervals of increase/decrease, constant interval - similar to [Activity 2.3.23](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/FN3.html#find-all-characteristics)

StevenClontz:
Closes #160

Closes #169

### DIFF
--- a/source/precalculus/exercises/outcomes/FN/FN3/generator.sage
+++ b/source/precalculus/exercises/outcomes/FN/FN3/generator.sage
@@ -1,10 +1,86 @@
+import numpy as np
 class Generator(BaseGenerator):
-    def data(self):
+  def data(self):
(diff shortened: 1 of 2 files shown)


siwelwerd:
Why do these other ones need reimplementing?  #160 looked like it cleanly changed the two necessary files and nothing else?

PR #256, opened 2024-08-23
Precalculus EL2 Checkit

### LINKED ISSUE #182: EL2 CheckIt
Model problem after [Activity 5.2.10](https://teambasedinquirylearning.github.io/precalculus/2024/instructor/EL2.html#activity-exp-graphs-fluency).

**Directions:** Graph the following exponential function, labeling at least two points on the graph. Include any asymptotes with a dotted line. State the domain, the range, and the equation of the asymptote.

The base should be one of the following:
- an integer between 2 and 6
- a simple fraction (like 1/n for n between 2 and 6)
- e

Function should include a reflection (x or y axis) and either a horizontal or vertical transformation. Like 

- $-3^{x-4}$
- $3^{-x}-4$
- $-3^x - 4$. 

Exclude ones like $3^{-x-4}$ with horizontal shift and reflection together.

siwelwerd:
Closes #143
Closes #182

### DIFF
--- a/source/precalculus/exercises/outcomes/EL/EL2/generator.sage
+++ b/source/precalculus/exercises/outcomes/EL/EL2/generator.sage
@@ -1,10 +1,45 @@
 class Generator(BaseGenerator):
-    def data(self):
-      return {
-        "eq": 3,
-        "ineq": 4,
-        "eqansinterval": 5,
-        "eqansgraph": 6,
-        "ineqansinterval": 7,
-        "ineqansgraph": 8
-      } 
+  def data(self):
+    a=1
+    b = choice([2..6]+[e])
+    c=0
+    d=0
+    k=1
+
+    reflection = choice(["vertical", "horizontal"])
+    if reflection == "horizontal":
+      shift="vertical"
+    else:
+      shift = choice(["vertical","horizontal"])
+    
+    if reflection == "horizontal":
+      k = -1
+    if reflection == "vertical":
+      a=-1
+    if shift == "horizontal":
+      c = choice([-4..-1,1..4])
+    if shift == "vertical":
+      d = choice([-5..-1,1..5])
+    
+    f=a*b^(k*x-c)+d
+
+    return {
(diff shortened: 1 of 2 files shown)


StevenClontz:
I added https://github.com/TeamBasedInquiryLearning/library/wiki/Style-Guide#sagemath-code but feel free to push back on that (particularly if you've found a good SageMath extension for VS Code).

StevenClontz:
@AbbyANoble requested `e` as an option but I removed it - if you're only asking it on 13% of seeds, I'm not sure you're really assessing it (and are just asking for "gotchas" when you ask some students about it but not others). Should we have a second part that always uses `e`?

StevenClontz:
merging without e for now, but can follow up with a new PR to implement it

StevenClontz (approved):

AbbyANoble:
> @AbbyANoble requested e as an option but I removed it - if you're only asking it on 13% of seeds, I'm not sure you're really assessing it (and are just asking for "gotchas" when you ask some students about it but not others). Should we have a second part that always uses e?

For what it's worth, CheckIt never directly assesses my students. Getting it to play right with D2L, being reasonably confident students are doing their own work, and dealing with a different question for every student are not worth the issues they create. In my world, CheckIt is a homework set for students, and in that situation I'm okay with 13% of them showing up with an e.  I use CheckIt to generate problems that I put on my own assessments I've created, usually a handful of similar versions rather than different for every student.

All that to say, to find that middle area, I would suggest making a graphing problem with e its own question.

StevenClontz:
I want CheckIt to produce problems appropriate for use both as homework and for assessing students on our learning outcomes across the library, even if that's not how it's used by some instructors (it works *really well* for me in Linear Algebra and Canvas or paper). I'll make an issue to add e back to this.

AbbyANoble:
Yeah. I get it. I just think more homework than assessment when I'm writing things. Gotta get out of that habit.

PR #274, opened 2024-08-27
remove mistake in LE2 activity, add doenet interactive for adding rows

### LINKED ISSUE #19: Interactive Geometry for LE2
Documenting a great suggestion from @owensks in our meeting today: LE2 would be enhanced by an interactive visualization showing how the different row moves may change a system but preserves their intersection points. Even a 2-D visualization would likely be sufficient here.

This is related to TeamBasedInquiryLearning/linear-algebra#165 , which is an old issue from when this did not seem easy to do.  But now there are tools like Doenet that make this feasible.

StevenClontz:
Closes #19

### DIFF
--- a/source/linear-algebra/source/01-LE/02.ptx
+++ b/source/linear-algebra/source/01-LE/02.ptx
@@ -109,9 +109,9 @@
 <activity estimated-time='10'>
     <introduction>
         <p>
-Consider whether these matrix manipulations (A) <em>must keep</em> or
-(B) <em>could change</em>
-the solution set for the corresponding linear system.
+Consider whether these matrix manipulations (A) must keep the <em>same</em>
+solution set, or (B) might result in a <em>different</em>
+solution set for the corresponding linear system.
         </p>
     </introduction>
     <task>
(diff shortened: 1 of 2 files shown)


source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -168,21 +182,28 @@ Add a constant to every term of a row, for example:
 <sidebyside>
     <p>
         <me>
-            \left[\begin{array}{cc|c}  1 &amp; 2 &amp; 3 
-                                    \\ 4 &amp; 5 &amp; 6 \end{array}\right]
+            \left[\begin{array}{cc|c}  1 &amp; 2 &amp; 4 
+                                    \\ 1 &amp; 3 &amp; 5 \end{array}\right]
         \sim
-            \left[\begin{array}{cc|c} 1+6 &amp; 2+6 &amp; 3+6
-                                    \\ 4 &amp; 5 &amp; 6 \end{array}\right]
```
StevenClontz, line 175: Actually adding a constant to rows in this example is perfectly fine (since the solution satisfies $x+y=1$ lol). 🤦‍♂️

New version does not satisfy $x+y=1$.

source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -192,44 +213,58 @@ Multiply a row by a nonzero constant, for example:
...
+    </ol>
+</p>
         </statement>
+        <answer>A. Same</answer>
     </task>
     <task>
         <statement>
             <p>
-Add a constant multiple of one row to another row, for example:
+Subtract one row by another row, for example:
```
StevenClontz, line 242: I wanted to make this a little easier than the general case, and make an interactive for the general case.

source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -264,24 +306,52 @@ Replace a row with zeros, for example:
...
     </task>
 </activity>
 
+<activity xml:id="LE2-activity-add-rows">
+    <statement>
+        <interactive label="AT1-interactive-add-rows" platform="doenetml" width="100%">
+          <slate surface="doenetml">
+            <xi:include parse="text" href="doenet/LE2-add-rows.xml"/>
+          </slate>
+        </interactive>
```
StevenClontz, line 340: image

StevenClontz:
Direct link: [Activity 1.2.3](https://e5d9360c.tbil.pages.dev/preview/linear-algebra/instructor/LE2#LE2-4-3)

source/linear-algebra/source/01-LE/02.ptx
siwelwerd: I think keeping it adding might be cleaner.

StevenClontz: Done in 4810968

jkostiuk (approved):
I really like this interactive, and think the corrections are great.

StevenClontz:
Hmm, this caused a bug in the PDF build. 😔 We probably need a staging branch to send PRs to that run the full build before deploying to main...

source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -122,19 +122,26 @@ Swapping two rows, for example:
...
     </p>
 </sidebyside>
+<p>
+    <ol marker="A." cols="2">
+        <li>Solutions must be the <em>same</em>.</li>
+        <li>Solutions might be <em>different</em>.</li>
+    </ol>
+</p>
         </statement>
+        <answer>A. Same</answer>
```
StevenClontz, line 144: Forgot to scaffold with `<p/>` 🤦‍♂️

siwelwerd:
> Hmm, this caused a bug in the PDF build. 😔 We probably need a staging branch to send PRs to that run the full build before deploying to main...

Can we* configure it to have an approval trigger a full build on a PRl?  And then only merge if that check passes?

*You

siwelwerd:
> Can we* configure it to have an approval trigger a full build on a PR? And then only merge if that check passes?

Looks like yes? https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-a-workflow-when-a-pull-request-is-approved

PR #275, opened 2024-08-27
scaffolding for LE1 activities 12 and 13

StevenClontz:
Adds some more scaffolding, e.g. MCQs, for LE1 activities. N.B. the first activity and change remain whiteboard activities to set the right tone, but MCQs are then introduced to clarify the requested tasks and avoid a few red herrings (though I did have some good discussion in class when a team suggested $0=0$ was the solution to the system).

Direct link [Activity 1.1.13](https://a888658b.tbil.pages.dev/preview/linear-algebra/instructor/LE1#LE1-4-13)

### DIFF
--- a/source/linear-algebra/source/01-LE/01.ptx
+++ b/source/linear-algebra/source/01-LE/01.ptx
@@ -781,37 +781,6 @@ Augmented matrix:
         </p>
         </statement>
     </example>
-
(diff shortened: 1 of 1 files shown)


siwelwerd (approved):

StevenClontz:
Hmm, this caused a bug in the PDF build. 😔 We probably need a staging branch to send PRs to that run the full build before deploying to main...

PR #285, opened 2024-08-28
Tweak precalc FN4 checkit

### LINKED ISSUE #281: FN4 CheckIt
- For reflections, change the language to "reflected over the x-axis" or "reflected over the y-axis" instead of "flipped". 
- Could the two graphs be shown less zoomed in? And on graphs with the same  range of x's and y's so it's more clear the changes that happened? Without paying attention to the scales closely, the horizontal and vertical shifts are not very apparent.
- And, this may be a general thing more than just this particular one, but the scaling of the axes to fit sort of loses perspective of where things are on the coordinate plane. I think seeing a more standard -10,10 (or -8 to 8) for the x's and y's would be better.

siwelwerd:
Closes #281 

@AbbyANoble does this look better?

image
image

### DIFF
--- a/source/precalculus/exercises/outcomes/FN/FN4/template.xml
+++ b/source/precalculus/exercises/outcomes/FN/FN4/template.xml
@@ -25,7 +25,7 @@
             <p>
                 To produce <m>{{gname}}(x)</m>, the graph of <m>{{fname}}(x)</m>is shifted {{hshift}} by <m>{{b}}</m>,
                 <!--{{#flip}}-->
-                flipped vertically,
+                reflected across the <m>y</m>-axis,
                 <!--{{/flip}}-->
                 and then shifted {{vshift}} by <m>{{c}}</m>.
             </p>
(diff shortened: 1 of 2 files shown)


AbbyANoble:
Yes! But that's a reflection across the x-axis, not y.

source/precalculus/exercises/outcomes/FN/FN4/template.xml
```diff
@@ -25,7 +25,7 @@
             <p>
                 To produce <m>{{gname}}(x)</m>, the graph of <m>{{fname}}(x)</m>is shifted {{hshift}} by <m>{{b}}</m>,
                 <!--{{#flip}}-->
-                flipped vertically,
+                reflected across the <m>y</m>-axis,
```
siwelwerd, line 28:
```suggestion
                reflected across the <m>x</m>-axis,
```

siwelwerd:
Github really needs to implement the 🤦🏻‍♂️ emoji as a reaction.

AbbyANoble:
It was the y-values that changed though. I'm sure that's what you were thinking of. 😁

AbbyANoble:
Hey @StevenClontz! Can you review this and the other CheckIt pull requests so they are updated (and I can use them next week!)?

StevenClontz:
Actually I need to make it so *you* can review these! I'm going ahead and merging these so you can look at them in the morning yourself; report back in a new Issue if there's any concerns.

StevenClontz (approved):

PR #294, opened 2024-09-03
LE3 updates

StevenClontz:


source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -811,6 +811,19 @@ are available online.
...
+    </p>
+    <program>
+        <input>
+A = [1 3 2
+     2 5 7]
+
+rref(A)
+        </input>
+    </program>
+</li>
```
StevenClontz, line 826: Explicitly demonstrate whitespace use with Octave RREF

source/linear-algebra/source/01-LE/03.ptx
```diff
@@ -59,6 +59,9 @@ Consider the following system of equations.
 </mrow>
 <mrow>
              -x_1 &amp;\,+\,&amp; 3x_2 &amp;\,-\,&amp;  6x_3 &amp;\,=\,&amp; 11
+</mrow>
+<mrow>
+             4x_1 &amp;\,+\,&amp; x_2 &amp;\,+\,&amp;  x_3 &amp;\,=\,&amp; 1
```
StevenClontz, line 64: Use 4 equations in each activity, mirroring quizzes.

source/linear-algebra/source/01-LE/03.ptx
```diff
@@ -234,16 +255,7 @@ How many solutions must this system have?
...
 </sage>
 
-<fact xml:id="linearsystemsfact2">
-    <statement>
-        <p>
-We will see in <xref ref="LE4"/> that the intuition established here generalizes: a <em>consistent</em> system with more variables
-than equations (ignoring <m>0=0</m>) will always have infinitely many solutions.
-        </p>
-    </statement>
-</fact>
```
StevenClontz, line 244: This felt a bit duplicative with the following Fact. But @jkostiuk should chime in - it's not clear to me what you're asking for in your Additional Practice activity that cites this (which I updated to cite the next fact).

source/linear-algebra/source/01-LE/03.ptx
```diff
@@ -253,27 +265,51 @@ we can immediately tell how many solutions the system has.
...
+<me>0=1</me>, that is, the <m>\RREF</m> matrix includes the row
+<me>\left[\begin{array}{ccc|c}0&amp;\cdots&amp;0&amp;1\end{array}\right]</me>,
 then the system is <em>inconsistent</em>, which means it has <em>zero</em> solutions
-and its solution set is written as <m>\emptyset</m> or <m>\{\}</m>.
+and we may write
+<me>
+    \text{Solution set }=\{\}
+    \hspace{2em}\text{or}\hspace{2em}
+    \text{Solution set }=\emptyset.
+</me>
```
StevenClontz, line 276: Make explicit solution vs. solution set.

source/linear-algebra/source/01-LE/03.ptx
```diff
@@ -253,27 +265,51 @@ we can immediately tell how many solutions the system has.
...
+</md>
+(with some possible extra <m>0=0</m> equations),
+then the system is <em>consistent</em> with exactly <em>one</em> solution,
+and we may write
+<me>
+    \text{Solution }=\left[\begin{array}{c}s_1\\s_2\\\vdots\\s_n\end{array}\right]
+    \hspace{2em}\text{but}\hspace{2em}
+    \text{Solution set }=
+        \left\{\left[\begin{array}{c}s_1\\s_2\\\vdots\\s_n\end{array}\right]\right\}.
+</me>
```
StevenClontz, line 306: Make explicit solution vs solution set

source/linear-algebra/source/01-LE/03.ptx
```diff
@@ -253,27 +265,51 @@ we can immediately tell how many solutions the system has.
...
           </li>
           <li>
               <p>
-Otherwise, the system must have more variables than non-trivial
-equations (equations other than <m>0=0</m>). This means it is
+Otherwise, the system given by the <m>\RREF</m> matrix must <em>not include a <m>0=1</m> contradiction</em>
+while <em>including at least one equation with multiple variables</em>. This means it is
 <em>consistent</em> with <em>infinitely-many</em>
 different solutions. We'll learn how to find such solution sets in 
 <xref ref="LE4"/>.
```
StevenClontz, line 315: I think we overcomplicated it previously: we have $\infty$-many solutions as long as the RREF has no contradictions, and at least one equation with multiple variables. No need to compare count of non-trivial equations vs variables: I just rule out 0 or 1 solution, then know it has $\infty$-many.

source/linear-algebra/source/01-LE/03.ptx
```diff
@@ -283,85 +319,86 @@ different solutions. We'll learn how to find such solution sets in
...
-  <introduction>
-    <p>
-For each vector equation, write an explanation for
-whether each solution set
-has no solutions, one solution, or infinitely-many solutions.
-If the set is finite, describe it using set notation.
+<activity>
+    <introduction>
+      <p> Consider each of the following systems of linear equations or vector equations. </p>
+    </introduction>
```
StevenClontz, line 325: Updated fluency builder to match LE3 quiz

source/linear-algebra/source/01-LE/03.ptx
jkostiuk: ya, i don't disagree and I like how the fact is stated now a bit better. I think I might have to think of a better individual practice for the future.

jkostiuk (approved):
based on what I saw in the preview, i like these changes.

jkostiuk:
@StevenClontz didn't know if a new issue was needed or a comment here but: given this change to LE3, Activity 1.4.2 should be adjusted to add two rows of zeros to RREF in (c)

siwelwerd:
General workflow is that once a PR is merged into main (like this one is), any new proposed changes need a new issue and/or PR.  If it's still open, we could continue the conversation and keep changing things.

PR #298, opened 2024-09-04
Consistency in using variable

siwelwerd:
`scenario_alt` is a variable containing strings such as "scenarioA".  Instead of testing e.g. `scenario_alt[-1]=="A"`, consistently test e.g. `scenario_alt=="scenarioA"` which is more readable.


source/precalculus/exercises/outcomes/EQ/EQ2/generator.sage
```diff
@@ -93,8 +93,8 @@ class Generator(BaseGenerator):
             }
         
         if scenario == "mixture_percent":
-            if scenario_alt[-1] in ["A", "B"]:
-                if scenario_alt == "A":
```
siwelwerd, line 97: This was a bug causing this block to never be executed.

StevenClontz (approved):

PR #307, opened 2024-09-10
Stub out trig modules

siwelwerd:


siwelwerd:
Self merging as this is blocking ongoing work on the Precalc book.

PR #309, opened 2024-09-11
update LE samples

StevenClontz:


StevenClontz:
Being bad and self-merging as I need this soon (I need to reinforce that students need to answer each task specifically).

@siwelwerd @jkostiuk feel free to call me out if there are concerns here.

PR #313, opened 2024-09-17
Prompt for a bit more detail in EV1 linear combo construction

StevenClontz:
I'm getting a lot of students who (I think) are guess-checking to a valid combination; this would better prompt them to learn how to use the matrix to extract the combo.

### DIFF
--- a/source/linear-algebra/exercises/outcomes/EV/EV1/template.xml
+++ b/source/linear-algebra/exercises/outcomes/EV/EV1/template.xml
@@ -46,8 +46,8 @@ The vector equation has no solutions, so <m>{{v}}</m> is not a linear combinatio
         <knowl>
             <content>
                 <p>
-If your statement was true, describe a linear combination of <m>{{veclist}}</m>
-that equals <m>{{v}}</m>.
+If your statement was true, describe how to use your RREF matrix to construct
+a linear combination of <m>{{veclist}}</m> that equals <m>{{v}}</m>.
                 </p>
             </content>
             <outtro>
--- a/source/linear-algebra/exercises/outcomes/EV/EV1/template.xml
+++ b/source/linear-algebra/exercises/outcomes/EV/EV1/template.xml
@@ -46,8 +46,8 @@ The vector equation has no solutions, so <m>{{v}}</m> is not a linear combinatio
         <knowl>
             <content>
                 <p>
-If your statement was true, describe how to use your RREF matrix to construct
-a linear combination of <m>{{veclist}}</m> that equals <m>{{v}}</m>.
+If your statement was true, explain and demonstrate how to construct
+a specific linear combination of <m>{{veclist}}</m> that equals <m>{{v}}</m>.
                 </p>
             </content>
             <outtro>
--- a/source/linear-algebra/source/02-EV/samples/01.ptx
+++ b/source/linear-algebra/source/02-EV/samples/01.ptx
@@ -1,214 +1,98 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <example xml:id="sample-EV1"><title>EV1</title>
-    <statement>
-        <ol>
-            <li>
+    <introduction>
+      <p>Consider each of these claims about a vector equation.</p>
+    </introduction>
+      <task>
+        <introduction>
+          <p><m>\left[\begin{array}{c} -13 \\ 3 \\ -15 \end{array}\right]</m>
+            is a linear combination of the vectors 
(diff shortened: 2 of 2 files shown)


siwelwerd:
>I'm getting a lot of students who (I think) are guess-checking to a valid combination

I think that's a valid solution strategy, and probably shows more understanding than memorizing something about an RREF matrix (how many of them can explain *why* those numbers in the RREF matrix describe the coefficients needed?)

StevenClontz:
> (how many of them can explain why those numbers in the RREF matrix describe the coefficients needed

Well, I thought I was suggesting we now ask them exactly this!

source/linear-algebra/exercises/outcomes/EV/EV1/template.xml
```diff
@@ -46,8 +46,8 @@ The vector equation has no solutions, so <m>{{v}}</m> is not a linear combinatio
         <knowl>
             <content>
                 <p>
-If your statement was true, describe a linear combination of <m>{{veclist}}</m>
-that equals <m>{{v}}</m>.
+If your statement was true, describe how to use your RREF matrix to construct
```
StevenClontz, line 49:
```suggestion
If your statement was true, explain and demonstrate how to use your RREF matrix to construct
```

siwelwerd:
I am guessing that in response to that prompt you would see a lot of "The numbers in column 4 give the coefficients to construct the the 4th vector as as a linear combination of the first 3" and then verifying that combination.  But that's not explaining at all why those numbers show up in that column.

I think you might have to ask them to write the solution set to get them to that level of reasoning.

Also, whatever change you want to make here needs to be reflected in the sample solution in the appendix.

StevenClontz:
What's your take @jkostiuk ?

StevenClontz:
FWIW my explanation would not yield a full solution set, but instead say "by substituting 0 into the free (non-pivot) variables, we obtain the equations x=X, y=Y, and w=W, and we may verify X[vect]+Y[vect]+0[vect]+W[vect]=[vect]". But of course I'd also be happy with a solution set, then plugging into free variables there.

Being able to guess and check is good, but avoiding cognitive overload each time you need a linear combo from an RREF is also good.

jkostiuk:
I'm largely neutral here.

In my experience using this assessment, I find most students do write out the solution set and then set different variables. Some of them use 0s, but others use 1s, and yet others use something more exotic or an educated guess-and-check. I actually like the variety that students take to the given prompt. 

If I had to vote one way or the other though, I think I'd vote to leave it---because I like that it's providing variety of reasoning as written for my students.

StevenClontz:
Are your students volunteering their reasoning? Because my students are mostly just writing a (usually) valid linear combination without further detail. So, unless I ask, I don't know their thought process.

Is there a way we can prompt better to get some level of detail beyond just writing a valid linear combination, that specifically reminds students to leave some artifact of their reasoning on the page?

jkostiuk:
Ya, they'll say something like "after trial and error I got blah" or "writing out the solutions set and setting a=blah", we get (blah). 

This is largely an artifact of us having established the understanding that we should always share our reasoning in such assessments and then practicing the checkit problems/looking at the template solutions.

StevenClontz:
I'm able to get there with my students, but the conversation is always way easier if I can point to where it specifically asked for the extra detail in the prompt. Hence all the "explain and demonstrates" rather than "what is" questions in the bank.

source/linear-algebra/exercises/outcomes/EV/EV1/template.xml
```diff
@@ -46,8 +46,8 @@ The vector equation has no solutions, so <m>{{v}}</m> is not a linear combinatio
...
+a linear combination of <m>{{veclist}}</m> that equals <m>{{v}}</m>.
```
StevenClontz, line 50:
```suggestion
If your statement was true, explain and demonstrate how to construct
a specific linear combination of <m>{{veclist}}</m> that equals <m>{{v}}</m>.
```

Maybe an okay compromise @jkostiuk @siwelwerd ?

jkostiuk: I like it!

StevenClontz:
Ah, still not ready until I update the practice exercises

StevenClontz:
I think this has approval so going ahead and self-merging.

jkostiuk:
Oh sorry, I got confused by your previous comment: I thought I was waiting on one more update from you before I was supposed to review/approve.

StevenClontz:
Sorry, I marked it was ready to review but should have commented explicitly.

PR #314, opened 2024-09-17
Prompt use of technology in LE2

StevenClontz:
For whatever reason I have a student who's stubborn about not using technology for RREF. Making sure students can use tech to compute RREF is probably at least as important as having them able to compute it by hand.


jkostiuk (approved):
I think this is a great suggestion

PR #320, opened 2024-09-18
Rename non-preview edition of Precalculus

siwelwerd:
Calling the 2024 edition the '2024 Preview edition' keeps causing confusion with the 'PREVIEW' edition.


StevenClontz (approved):

StevenClontz:
Thanks - couldn't think of a good synonym for preview. Trial Edition is great.

PR #342, opened 2024-10-07
Fix LF graph scales

### LINKED ISSUE #340: Graphs in LF1 and LF2 checkit
The graphs in LF1 Task 2.3 need to be formatted differently in a few ways.

1. First, so it's easier to see what points are on the graph in some of them. Maybe it's my old eyes, but it's hard for me to tell for sure what the whole number points are. For instance, in this one below (which I generated from the assessment builder), I thought it went through (0,-4) and (1,-1), giving a slope of 3. But looks like I'm off.

Not sure of the best fix, but could include making the line thinner, zooming in to a more standard window, then enlarging it some.
image

The next one I generated looked like this, which was easier to read:

image

2. I'd also suggest making sure the origin is in every graph. See how the y-axis in the one below (also generated) goes from -3 to -9. I think that may cause extra confusion in this course, and it just doesn't look very standard. 

image

3. When the line is horizontal, it really zooms in. I'd guess this happens for vertical ones too, but that hasn't popped up while I've been clicking re-generate.

image

Ultimately, I think a more standard size and domain/range of -10 to 10 type of window may help a lot of these issues. And be more familiar for students at this level.

siwelwerd:
Closes #340

### DIFF
--- a/source/precalculus/exercises/outcomes/LF/LF1/generator.sage
+++ b/source/precalculus/exercises/outcomes/LF/LF1/generator.sage
@@ -62,5 +62,5 @@ class Generator(BaseGenerator):
 
     @provide_data
     def graphics(data):
-       p=plot(data["g"],-6,6,thickness=3,gridlines=True,ticks=[1,1],aspect_ratio=1)
+       p=plot(data["g"],-10,10,ymin=-10,ymax=10,thickness=2,gridlines=True,ticks=[1,1],aspect_ratio=1)
        return {"plot":p}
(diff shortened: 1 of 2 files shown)


AbbyANoble:
I'm not sure how to review a CheckIt change since I can't preview. I looked at the code, but I don't know enough there to make a decision.

siwelwerd:
Unfortunately changes like this aren't easily captured in a screenshot, someone needs to flip through a dozen or so of them to get a feel :(

Press `,` to fire up a codespace from this branch (or open a codespace otherwise and switch to this branch).  Then you just need to run a line of code in the terminal (which [is in our Wiki](https://github.com/TeamBasedInquiryLearning/library/wiki/CheckIt-Authoring))

`python scripts/preview_outcome.py precalculus LF1`

When it's done it should give you a link that will open a new tab with the preview of that outcome.

AbbyANoble (approved):

PR #347, opened 2024-10-10
Adjust EV6 spanning set

StevenClontz:
Grading some EV6, and I don't like the examples of three vectors spanning dimension two: I'd rather each problem have a bit more meat to it.

Likewise edited to ensure the vectors don't span all of R^4, though I've yet to have any student tell me that {e_1,e_2,e_3,e_4} is a basis in that case.

### DIFF
--- a/source/linear-algebra/exercises/outcomes/EV/EV6/generator.sage
+++ b/source/linear-algebra/exercises/outcomes/EV/EV6/generator.sage
@@ -3,15 +3,15 @@ load("../sage/common.sage")
 class Generator(BaseGenerator):
     def data(self):
         #Pick some  vectors in R4
-        n=choice([3,4,5])
-        dim=choice(range(2,n))
+        n=choice([4,5])
+        dim=choice([2,3])
         A=CheckIt.simple_random_matrix_of_rank(dim,columns=n,rows=4)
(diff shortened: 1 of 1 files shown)


StevenClontz:
(Thanks for the tags Drew.)

Jordan I don't have the CheckIt previewing live here, but if you trust me on the code, I'm looking for a content review of changing every https://tbil.org/preview/linear-algebra/exercises/#/bank/EV6/1/ to have either 4 or 5 vectors, and the solution to always be either dimension 2 or 3.

siwelwerd:
Instructions for previewing Checkit problems reside [in our Wiki](https://github.com/TeamBasedInquiryLearning/library/wiki/CheckIt-Authoring)

And FWIW I eyeballed the code and it looks fine, but @jkostiuk can look at the outputted problems and verify.

jkostiuk (approved):
this looks good to me. Thanks @siwelwerd for the CheckIt info--I was able to follow the instructions and click a few exercises. 

@StevenClontz once this is merged, will it update the public-facing CheckIt bank?

StevenClontz:
It will update https://tbil.org/preview/linear-algebra/exercises/#/bank/EV6/1/ but not the 2024 edition.

Issue #352, opened 2024-10-11
EV7 Suggestion: More complicated example


jkostiuk:
In my opinion, the examples in EV7 are not complex enough to be convincing that the bases that are obtained are actually linearly independent. 

I'm going to open this issue for discussion but also share a suggestion for a change.

PR #353, opened 2024-10-11
Issue 352-EV7-suggestion

jkostiuk:
My attempt at addressing #352. 

First, I got rid of part (c) in 2.7.3 and changed the MC options in the revised activity. My thinking is that this more focused now around the (in)dependence and spanning/not spanning of the solution set. 

Second, I copied this sequence with a more complex example (this time having 4 free variables). 

My reasoning is that this example is now meaty enough that we will be forced to consider how we might test that the resulting set is independent. I suspect some teams may defer to the methods of EV4/EV6 and then we probe them about why it's the case that these guys are automatically independent. 

Because this example is so meaty, I got rid of the two examples that follow but kept the one where the solution space is trivial.


jkostiuk:
tagging @StevenClontz @siwelwerd @jford1906 for thoughts

source/linear-algebra/source/02-EV/07.ptx
```diff
@@ -128,13 +128,73 @@ Rewrite this solution space in the form <me>\setBuilder{ a \left[\begin{array}{c
 </task>
 <task>
     <p>
-Rewrite this solution space in the form <me>\vspan\left\{\left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown\end{array}\right], \left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown \end{array}\right]\right\}.</me>
+Which of these choices best describes the set of two vectors
+<m>\left\{\left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown\end{array}\right], \left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown \end{array}\right]\right\}</m>
+used in this span?
```
StevenClontz, line 133:
```suggestion
used in this solution space?
```

source/linear-algebra/source/02-EV/07.ptx
```diff
@@ -128,13 +128,73 @@ Rewrite this solution space in the form <me>\setBuilder{ a \left[\begin{array}{c
...
+    <p>
+        The set fails to span the solution space.
+    </p>
+  </li>
+  <li>
+    <p>
+        The set is a basis of the solution space. 
+    </p>
+  </li>
+</ol>
```
StevenClontz, line 160: Generally, I prefer to keep things to 4 choices (in case instructor's multiple choice mechanism only supports A-D, e.g. the flashcards I have in my classroom).

source/linear-algebra/source/02-EV/07.ptx
```diff
@@ -128,13 +128,73 @@ Rewrite this solution space in the form <me>\setBuilder{ a \left[\begin{array}{c
...
+x_1&amp;\,+\,&amp;2x_2&amp;\,+\,&amp;x_3 &amp;\,+\,&amp; x_4 &amp;\,+\,&amp;3x_5&amp;\,+\,&amp;6x_7&amp;\,+\,&amp;7x_7&amp;=&amp; 0
+    </mrow>
+</md>
+        </p>
+    </introduction>
+<task>
+    <p>
+Find its solution set (a subspace of <m>\IR^4</m>).
+    </p>
+</task>
```
StevenClontz, line 188: Since this is such a hefty example, can we just provide the solution set?

StevenClontz:
While we're touching this, can we add `<answer>`s for the instructor edition?

StevenClontz:
Last comment for now: maybe in Fact 2.7.5 we should `\textcolor` these rows to emphasize why RREF of the appropriate matrix will have all pivot columns:

image

Converting this to a draft while we discuss things.

source/linear-algebra/source/02-EV/07.ptx
```diff
@@ -128,13 +128,73 @@ Rewrite this solution space in the form <me>\setBuilder{ a \left[\begin{array}{c
...
     </p>
 </task>
 <task>
     <p>
 Which of these choices best describes the set of two vectors
-<m>\left\{\left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown\end{array}\right], \left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown \end{array}\right]\right\}</m>
+<m>\left\{\left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown\\ \unknown\\ \unknown \\ \unknown\end{array}\right], \left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown\\ \unknown\\ \unknown \\ \unknown\end{array}\right],\left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown\\ \unknown\\ \unknown \\ \unknown\end{array}\right],\left[\begin{array}{c} \unknown \\ \unknown \\ \unknown \\ \unknown\\ \unknown\\ \unknown \\ \unknown\end{array}\right]\right\}</m>
 used in this span?
```
StevenClontz, line 198:
```suggestion
used in this solution space?
```

source/linear-algebra/source/02-EV/07.ptx
jkostiuk:
```suggestion
<ol marker="A.">
<li>
    <p>
    The set is linearly dependent.
    </p>
  </li>
<li>
    <p>
        The set is linearly independent.
    </p>
  </li>
  <li>
      <p>
        The set spans the solution space.
      </p>
    </li>
  <li>
    <p>
        The set is a basis of the solution space. 
    </p>
  </li>
</ol>
```

source/linear-algebra/source/02-EV/07.ptx
jkostiuk: in my experience, this is good practice. Each term I've had students make up their own 3x7 matrix and do the procedure to get practice. They almost always do it correctly and troubleshoot the errors, even with messy fractions and such. By giving them this specific example now, I'd be able to use the teams to troubleshoot since they'll all be getting the same answer.

jkostiuk:
> While we're touching this, can we add `<answer>`s for the instructor edition?

do we give answers for all parts, or just the MC?

jkostiuk:
> Last comment for now: maybe in Fact 2.7.5 we should `\textcolor` these rows to emphasize why RREF of the appropriate matrix will have all pivot columns:
> 
> image
> 
> Converting this to a draft while we discuss things.

great idea

StevenClontz:
> > While we're touching this, can we add `<answer>`s for the instructor edition?
> 
> do we give answers for all parts, or just the MC?

Definitely MCQs, other things can't hurt either, with the ultimate goal of having answers for everything someday (at least enough for an instructor to use).

jkostiuk:
IMG_2023

leaving for the night, but here's the work for the example so we can add the answer and possibly consider restructuring the activity

source/linear-algebra/source/02-EV/07.ptx
StevenClontz: Sounds fine - if we have the answer in the instructor version, then an instructor can choose to spoil the solution set if they decide to move more quickly for pedagogical or time considerations.

siwelwerd:
>there's a conflict to merge...

It was the settings.json file, I just discarded those changes so it will merge fine.  You had started work on this before we de-tracked the settings.json file is why that popped up.

source/linear-algebra/source/02-EV/07.ptx
```diff
@@ -128,14 +128,74 @@ Rewrite this solution space in the form <me>\setBuilder{ a \left[\begin{array}{c
...
+  </li>
+</ol>
+    </p>
+</task>
+</activity>
+<answer>
+    <p>
+        D.
+    </p>
+</answer>
```
StevenClontz, line 163: The above `<task>` needs a `<statement>`, and then this `<answer>` should be a sibling of that statement.

jkostiuk:
@StevenClontz just made a commit--is that what you mean?2

source/linear-algebra/source/02-EV/07.ptx
```diff
@@ -149,22 +209,34 @@ used in this span?
...
   </li>
 </ol>
     </p>
 </task>
 </activity>
+<answer>
+    <p>
+        D.
+    </p>
+</answer>
```
StevenClontz, line 232: Similar here.

source/linear-algebra/source/02-EV/07.ptx
```diff
@@ -149,22 +209,34 @@ used in this span?
...
 
 <sage language="octave">
+    <input>
+    row1=[]
+    row2=[]
+    row3=[]
+    rref([row1;row2;row3])
+    </input>
 </sage>
```
StevenClontz, line 241: I think we usually just have a blank sage cell when we don't provide the matrix terms.

source/linear-algebra/source/02-EV/07.ptx
```diff
@@ -154,13 +155,16 @@ used in this solution space?
...
 </ol>
     </p>
+</statement>    
 </task>
-</activity>
 <answer>
     <p>
         D.
     </p>
 </answer>
```
StevenClontz, line 164: This still needs to live inside the `<task>` (cut-paste into line 159)

StevenClontz (approved):

Issue #354, opened 2024-10-11
AT6: Adding Activities to Find Bases for Matrix and Polynomial Spaces


jkostiuk:
Ever since we compartmentalized the section on abstract vector spaces, I've wanted to restructure some things in AT6 so that my students get practice actually verifying/finding bases for abstract vector spaces. 

We're going to be getting to this unit in a week or two, so I'm going to draft a suggestion quick for your input.

StevenClontz:
Can we rename this and #356 to describe the suggestions?

jkostiuk:
tried these in class today. I was really happy with them--I had a much better conversation about finding these bases than my previous attempts at improvising.

StevenClontz:
Same - Thanks for this work!

I think this is now closable; reopen if I'm overlooking anything.

PR #356, opened 2024-10-11
Adding Matrix and Polynomial Basis Activity to Address #354

jkostiuk:
Here is my suggestion for #354

AT6 relies upon us having bases for our vector spaces. I decided to structure the discussion around our vector-equations to help reinforce how "all we're doing" is using our old stuff in a new context. We can do this *if* we have a basis and so, first, we need to justify we have a basis.

### DIFF
--- a/source/linear-algebra/source/03-AT/06.ptx
+++ b/source/linear-algebra/source/03-AT/06.ptx
@@ -391,7 +391,7 @@ S=\setList{1,y,y^2,y^3,y^4}.
     </task>
     <task>
         <statement>
-            <p> What is the dimension of <m>\P^5</m>?
+            <p> What is the dimension of <m>\P^4</m>?
                 <ol marker="A." cols="2">
                     <li>2</li>
                     <li>3</li>
--- a/source/linear-algebra/source/03-AT/06.ptx
+++ b/source/linear-algebra/source/03-AT/06.ptx
@@ -165,7 +165,7 @@ Which of these completes the following definition for a bijective linear map
 <activity>
     <introduction>
         <p>
-Consider matrix space <m>M_{2,2}=\left\{\left[\begin{array}{cc}
+Consider the matrix space <m>M_{2,2}=\left\{\left[\begin{array}{cc}
     a&amp;b\\c&amp;d
 \end{array}\right]\middle| a,b,c,d\in\IR\right\}</m> and the following matrix equation:
 <me>
--- a/source/linear-algebra/source/03-AT/06.ptx
+++ b/source/linear-algebra/source/03-AT/06.ptx
@@ -162,31 +162,94 @@ Which of these completes the following definition for a bijective linear map
 </fact>
 
 
-
 <activity>
     <introduction>
         <p>
-The matrix space <m>M_{2,2}=\left\{\left[\begin{array}{cc}
+Consider matrix space <m>M_{2,2}=\left\{\left[\begin{array}{cc}
     a&amp;b\\c&amp;d
-\end{array}\right]\middle| a,b,c,d\in\IR\right\}</m> has the basis
+\end{array}\right]\middle| a,b,c,d\in\IR\right\}</m> and the following matrix equation:
 <me>
-\left\{
-\left[\begin{array}{cc}
+x_1\left[\begin{array}{cc}
     1&amp;0\\0&amp;0
-\end{array}\right],
-\left[\begin{array}{cc}
+\end{array}\right]+x_2\left[\begin{array}{cc}
     0&amp;1\\0&amp;0
-\end{array}\right],
-\left[\begin{array}{cc}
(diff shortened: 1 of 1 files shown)


jkostiuk:
@StevenClontz I'm tagging you since you're also teaching this upcoming. If we get this merged (possibly after some helpful suggestions), I'd be happy to report how this lesson goes (which I expect will be the week after next).

source/linear-algebra/source/03-AT/06.ptx
```diff
@@ -162,31 +162,94 @@ Which of these completes the following definition for a bijective linear map
 </fact>
 
 
-
 <activity>
     <introduction>
         <p>
-The matrix space <m>M_{2,2}=\left\{\left[\begin{array}{cc}
+Consider matrix space <m>M_{2,2}=\left\{\left[\begin{array}{cc}
```
StevenClontz, line 168:
```suggestion
Consider the matrix space <m>M_{2,2}=\left\{\left[\begin{array}{cc}
```

source/linear-algebra/source/03-AT/06.ptx
```diff
@@ -162,31 +162,94 @@ Which of these completes the following definition for a bijective linear map
...
+                        </me>
+                <ol marker="A." cols="2">
+                    <li>1</li>
+                    <li>infinitely many</li>
+                    <li>none</li>
+                    <li>2</li>
+                </ol>
+            </p>
+        </statement>
+    </task>
```
StevenClontz, line 223: image

Minor point: the equation in the introduction probably belongs in the first task.

Major point: how do you expect students to answer these questions? They cannot convert the equation to a Euclidean vector equation (that's the point of this activity, to convince them that they can convert to $\mathbb R^4$), which puts their usual RREF approach out of reach.

But what about this? Just ask if (the basis) is a spanning set, with answer choices:

- No, [example] is not a linear combination of these matrices. (3x)
- Yes, every matrix in M22 is a linear combination of these matrices.

Next part: is (the basis) linearly independent?

- No, [matrix] is a linear combination of [matrices] (3x)
- Yes, each matrix cannot be expressed as a linear combination of other matrices in the set.

Then you're set up for part (c) without having to solve equations.

StevenClontz:
I like the idea of having students develop the bases for $M_{2,2}$ and $\mathcal P_n$, but solving matrix/polynomial equations directly is exactly what conversion to Euclidean vectors is meant to avoid; I think we can establish these bases without having to manually solve such equations.

StevenClontz (changes requested):
I'm not yet convinced this the right scaffolding, but it's moving in a good direction.

source/linear-algebra/source/03-AT/06.ptx
jkostiuk: So, what I coach them to do in the past is to simplify the LHS of that matrix equation into $$\left[\begin{array}{cc}x_1&x_2\\x_3&x_4\end{array}\right]$$ and then we're just looking at $$\left[\begin{array}{cc}x_1&x_2\\x_3&x_4\end{array}\right]=$$ some specific matrix. 

My students seem to have a hard time really processing how to build a linear combination of matrices and so I think having the LHS written out gives them a starting point.

jkostiuk: ultimately, I think they'll be forced do what I describe (or take a more flexible approach) with your scaffolding, so I'm tempted to give it a shot. I also think your scaffolding would lead to more diversity of discussion in a productive way.

StevenClontz: Oh of course. I too overlooked the obvious approach to turn [matrix]=[matrix] into a *linear system* rather than trying to find the right Euclidean vector equation. So scaffolding in that direction works as well, but I think the made up constants on the right-hand side is distracting more than building on their intuition of "spanning means we can build everything" and "independence means there's no redundant information" to approach this.

jkostiuk: agreed. i've got a commit coming in with the suggested changes for your review

source/linear-algebra/source/03-AT/06.ptx
```diff
@@ -231,15 +311,87 @@ The matrix space <m>M_{2,2}=\left\{\left[\begin{array}{cc}
...
+                    <li><m>S</m> spans <m>\P^4</m></li>
+                    <li><m>S</m> is a basis of <m>\P^4</m></li>
+                </ol>
+            </p>
+        </statement>
+    </task>
+    <task>
         <statement>
-            <p> What is the dimension of <m>\P^4</m>?
+            <p> What is the dimension of <m>\P^5</m>?
```
StevenClontz, line 394:
```suggestion
            <p> What is the dimension of <m>\P^4</m>?
```

StevenClontz (approved):
I like it, thanks!

PR #358, opened 2024-10-15
Give compatible values for examples in AT2

StevenClontz:
Want to provide some values to actually verify for the class that each type of information is sufficient. Making a draft so I can expand into a full observation after my next class.

### DIFF
--- a/source/linear-algebra/source/03-AT/02.ptx
+++ b/source/linear-algebra/source/03-AT/02.ptx
@@ -196,17 +196,20 @@ What piece of information would help you compute <m>T\left(\left[\begin{array}{c
 <ol marker="A." cols="2">
 <li>
     <p>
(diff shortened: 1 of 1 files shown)


source/linear-algebra/source/03-AT/02.ptx
```diff
@@ -196,17 +196,20 @@ What piece of information would help you compute <m>T\left(\left[\begin{array}{c
 <ol marker="A." cols="2">
 <li>
     <p>
-    The value of <m>T\left(\left[\begin{array}{c} 0\\-4\\0\end{array}\right]\right)</m>.
+    The value of <m>T\left(\left[\begin{array}{c} 0\\-4\\0\end{array}\right]\right)
```
StevenClontz, line 199:
```suggestion
    The value of <m>T\left(\left[\begin{array}{c} 0\\-4\\0\end{array}\right]\right)=
```

jkostiuk (approved):

Issue #348, opened 2024-10-10
EV5: prompt for understanding about solutions AND spanning/independence


StevenClontz:
I'm grading https://library.tbil.org/2024/linear-algebra/exercises/#/bank/EV5/1/ and I wish we had this rigged to dig into students' understanding of not only "this equation has exactly one solution for every vector", but also explicitly connecting to their knowledge about spanning and independence.

I don't have any clever suggesitons for how to do so offhand, but noting this here for discussion.

jkostiuk:
[EV_5 (dragged).pdf](https://github.com/user-attachments/files/17346045/EV_5.dragged.pdf)

Here is what I've done for EV5 in the past. It was inspired because the sample solutions would address part (a) even though the checkit bank asked for part (b). Now I have them write the statement both ways.

Perhaps this is a starting point?

PR #359, opened 2024-10-15
update ev5 sample

StevenClontz:
Stopgap until #348 is implemented to make our sample match the current version of the bank.


jkostiuk (approved):
nice. i like the alternates too

PR #368, opened 2024-10-16
LF6 update

AbbyANoble:
Answer choices A and C were the same (6c + 4a vs 4a + 6c). The answer said A, so I assume C was supposed to be different and switched the variables to make what I think was the intended wrong answer. Tagging @kathypinzon as a reviewer since she was the original author!

Link to original: [Activity 3.6.2(b)](https://library.tbil.org/2024/precalculus/LF6.html#LF4-carnival-prices-3) 

Screenshot of original: 
image


source/precalculus/source/03-LF/06.ptx
```diff
@@ -55,7 +55,7 @@
           <li><p> <m>6c+4a=208</m> </p></li>
           <li><p> <m>a+c=208
           </m> </p></li>
-          <li><p> <m>4a+6c=208</m> </p></li>
+          <li><p> <m>4c+6a=208</m> </p></li>
```
StevenClontz, line 58: Would it be good (throughout this activity) to always list the `a` term first?

StevenClontz: Or vice/versa (children then adults), but either way, write things consistently to reduce cognitive overload of learners?

AbbyANoble: Consistency in order sounds good to me.

StevenClontz (changes requested):
As discussed, let's edit so variables appear in a consistent order.

source/precalculus/source/03-LF/06.ptx
AbbyANoble: About to make an update.

source/precalculus/source/03-LF/06.ptx
```diff
@@ -32,11 +32,11 @@
...
       <p>
         <ol marker="A." cols="1">
-          <li><p> <m>4a+2c=128.50</m> </p></li>
-          <li><p> <m>a+c=128.50
+          <li><p> <m>2c+4a=128.50</m> </p></li>
+          <li><p> <m>c+a=128.50
           </m> </p></li>
           <li><p> <m>4c+2a=128.50</m> </p></li>
-          <li><p> <m>a+c=336.50</m> </p></li>
+          <li><p> <m>c+a=336.50</m> </p></li>
```
StevenClontz, line 39: Should we add $ signs?

source/precalculus/source/03-LF/06.ptx
```diff
@@ -53,10 +53,10 @@
       <p>
         <ol marker="A." cols="1">
           <li><p> <m>6c+4a=208</m> </p></li>
-          <li><p> <m>a+c=208
+          <li><p> <m>c+a=208
           </m> </p></li>
-          <li><p> <m>4a+6c=208</m> </p></li>
-          <li><p> <m>a+c=336.50</m> </p></li>
+          <li><p> <m>4c+6a=208</m> </p></li>
+          <li><p> <m>c+a=336.50</m> </p></li>
```
StevenClontz, line 59: I think we want $ here too. (use `\$` in math mode)

StevenClontz:
Thanks @AbbyANoble, use the re-request review button when this is ready:

image

AbbyANoble:
> Thanks @AbbyANoble, use the re-request review button when this is ready:

Oops. I re-requested without reading all the way. 

I can go either way on the dollar signs. Personally, I feel like they junk up a system of equations.

StevenClontz:
In that case, do we need to remove them further down? [Activity 3.6.2(c)](https://f32358f8.tbil.pages.dev/preview/precalculus/instructor/LF6#LF4-carnival-prices-4)

AbbyANoble:
> In that case, do we need to remove them further down? [Activity 3.6.2(c)](https://f32358f8.tbil.pages.dev/preview/precalculus/instructor/LF6#LF4-carnival-prices-4)

Those don't bother me because they are an answer (where I would use units) instead of an abstraction of the problem. But I'm good with putting them everywhere. I'll throw them in.

StevenClontz (approved):
Maybe I'm a pedant, but I definitely prefer consistent use of units; I know our colleagues that apply mathematics to their fields appreciate it when we take care with dimensional analysis.

Issue #372, opened 2024-10-18
Calculus SQ3 Series Definition 


jbunn3:
This is something I'm pretty adamant about, so I have to bring this up. In [Definition 8.3.12](https://tbil.org/preview/calculus/SQ3.html#definition-SQ3serieslimit), we define the limit of a series (a number), and then turn around and define an infinite series to be that number. 

I will draft a PR on Monday to clarify the definition of a series as a pair of sequences and create a Remark about common uses (and abuses) of series notation. This change might require some modifications to the rest of the section to improve the flow.

StevenClontz:
Hm - my personal formaliation of a series is that it's a property of a sequence: the limit of its partial sums. I'm interested to see what your approach is.

PR #375, opened 2024-10-21
Modified Defn 8.3.12 for clarity and added a remark.

jbunn3:
For #372, This change clarifies the definition of a series and helps the careful reader distinguish between a series and its sum (limit).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,7 @@
     "xml.fileAssociations": [
         {
             "pattern": "**/source/**.ptx",
-            "systemId": "/home/vscode/.ptx/2.6.2/core/schema/pretext.rng"
+            "systemId": "/home/vscode/.vscode-remote/extensions/oscarlevin.pretext-tools-0.21.0/assets/schema/pretext.rng"
         }
     ]
 }
--- a/source/calculus/source/08-SQ/03.ptx
+++ b/source/calculus/source/08-SQ/03.ptx
@@ -251,10 +251,15 @@ More generally, partial sums may be defined for any starting index. Given <m>\{a
         <definition xml:id="definition-SQ3serieslimit">
             <statement>
                 <p>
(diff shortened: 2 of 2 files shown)


StevenClontz:
Defining a series as a pair of a sequence and its partial sum sequence doesn't click with me as a formalization: the latter is calculated from the former, so there's a redundancy that violates DRY (Don't Repeat Yourself).

But what if we say that a series is a *function* that maps a sequence to its partial sum sequence? I'll make a competing PR for discussion to elaborate on this.

siwelwerd:
>But what if we say that a series is a function that maps a sequence to its partial sum sequence?

I am wondering if getting this exactly precisely right in the main text is going to help our students, or if it belongs in an `<aside>` so as not to distract them from what we want them to learn.  In particular I worry that introducing a function to this definition is extra cognitive load.

StevenClontz:
The function idea didn't survive the drafting I did at #377. But you hit the nail on the head with this cognitive load concern: I don't think 99% of students are going to appreciate a model of a series that's a pair of sequences. I ended up saying that a series *is* a partial sum sequence.

jbunn3:
I agree with your thoughts that the cognitive load might not be right for the students. However, I am generally a fan of "planting a seed" that may help down the line, even if it doesn't make sense at the time. I say this since this is how I learned series as a student, and I didn't understand at that time either. However, later down the road, it made sense to me.

StevenClontz:
What value does having a series be a tuple of sequences have in contrast with just being the partial sum sequence?

StevenClontz:
Ah, I see your comment at #377, so I've marked it as ready for review, and it will close this PR once it is complete. Please suggest any improvements there. Thanks @jbunn3!

jbunn3:
> What value does having a series be a tuple of sequences have in contrast with just being the partial sum sequence?
I think it's mostly preference. I like the nature of thinking of them both back and forth. Yes, you can think of a partial sum sequence as a property of a sequence, but you can also think of the generating sequence as a property of a given partial sum sequence (by looking at the difference of consecutive elements).

PR #383, opened 2024-10-22
Ensure graphs include origin

### LINKED ISSUE #374: Graph scales in FN2
This is similar to #340.

Can we change the scales in Task 2 of FN2 the same way as LF1 and LF2 were fixed? Feels like these graphs are unnecessarily adding to the cognitive load here. I'd want to avoid things like this where the origin isn't shown in the graph:

Screenshot 2024-10-20 at 7 28 36 PM

Maybe a rescaling, counting by twos/fives or something to get to those numbers. 

And this one is especially nonstandard feeling: 
image

I don't mind the ones like this where the x's and y's have different ranges. But if it's easier to make them some standard size that's okay.
Screenshot 2024-10-20 at 7 31 29 PM

siwelwerd:
Closes #374


siwelwerd:
They should all include the origin now, even if they have to be zoomed out a bit like this one

image

AbbyANoble (approved):

PR #365, opened 2024-10-15
Tonya-10-15-2024

tdegeorge:

PR #385, opened 2024-10-22
Tonya 10 19 2024

tdegeorge:


siwelwerd:
@StevenClontz All but the last commit here seem to be from #365, which was already merged into main.  What's the best way to deal with that?  It seems like it shouldn't be showing a diff on 06-TR/03.ptx and 06-TR/04.ptx, for example, but it is.

StevenClontz:
I think a rebase from main will fix the diff, but will break @tdegeorge's Git history, so I'll make that a separate branch and PR. Marking this as draft to move discussion there.

StevenClontz:
That doesn't seem to be working. I'll just redo Tonya's work on a new branch.

PR #388, opened 2024-10-24
Cg2wilson ad3 patch 1

StevenClontz:
Fixes PR pointing to wrong branch


StevenClontz:
Actually it doesn't. @cg2wilson don't forget that new PRs need to be branched off of `main`. It's a simple fix so I'll reimplement on a new PR.

PR #397, opened 2024-11-04
Precalculus TE1

tdegeorge:


source/precalculus/source/08-TE/01.ptx
```diff
@@ -12,6 +12,1187 @@
     <subsection>
     <title>Activities</title>
 
+    <remark>
+      Pythagorean Identities, as we have seen in a previous section, are derived from the Pythagorean Theorem <m> \left( a^2+b^2=c^2 \right) </m>. Pythagorean Identities are used in solving many trigonometric problems where one trigonometric ratio is given and we are expected to find the other trigonometric ratios. The most commonly used Pythagorean Identity is: <me>\sin^2\theta + \cos^2\theta = 1</me>. The next two activities will lead us to find the other two Pythagorean Identities.
```
siwelwerd, line 16: We should link to the section and ideally the activity/observation/etc. this is done in (though I can't find where this is done previously).  

If needed, the subsequent activity can easily be rewritten to derive this.

source/precalculus/source/08-TE/01.ptx
```diff
@@ -12,6 +12,1187 @@
...
+          </p>
+        </answer>
+      </task>
+    </activity>
+
+    <remark>
+      In <xref ref="deriving-pythagorean-identity-2"/>, we derived the second Pythagorean Identity: <m>\sec^2\theta-\tan^2\theta=1</m>. Note that we can rewrite what we got in the previous activity <m> \left( 1 + \tan^2\theta = \sec^2\theta \right) </m> to get to this form. 
+    </remark> 
+
+    <activity xml:id="deriving-pythagorean-identity-3">
```
siwelwerd, line 149: Is it worth the time to go through all of this again (or a third time), or would it be easier to just tell them to take $\sin^2(x)+\cos^2(x)=1$ and divide it by $\sin^2(x)$, etc.?

source/precalculus/source/08-TE/01.ptx
```diff
@@ -12,6 +12,1187 @@
...
+          <p>
+            <m>\cos^2\theta=\frac{1+\cos2\theta}{2}</m>
+          </p>
+        </answer>
+      </task>
+
+      <task>
+        <statement>
+          <p>
+            To generate the power reduction formula for tangent, let's begin with its definition: <m>\tan^2\theta=\frac{\sin^2\theta}{cos^2\theta}</m>. Use the formulas you created in parts (a) and (b) to rewrite <m>\tan^2\theta</m>.
```
siwelwerd, line 951:
```suggestion
            To generate the power reduction formula for tangent, let's begin with its definition: <m>\tan^2\theta=\frac{\sin^2\theta}{\cos^2\theta}</m>. Use the formulas you created in parts (a) and (b) to rewrite <m>\tan^2\theta</m>.
```

source/precalculus/source/08-TE/01.ptx
```diff
@@ -12,6 +12,1187 @@
...
+      <introduction>
+        <p>
+          The next set of identities is the set of half-angle formulas, which can be derived from the reduction formulas. Let's derive the half-angle formula for <m>\sin\frac{\theta}{2}</m>.
+        </p>
+      </introduction>
+
+      <task>
+        <statement>
+          <p>
+            Recall the power reduction formula: <m>\sin^2\theta=\frac{1-\cos2\theta}{2}</m>. Replace <m>\theta</m> with <m>\frac{\theta}{2}</m>.
```
siwelwerd, line 1058: Would this be easier to have students set $\theta = \frac{\alpha}{2}$ so that $\theta$ isn't overloaded?  If not, then I think the following definition[sic] should also use $\theta$ to match the activity.

source/precalculus/source/08-TE/01.ptx
```diff
@@ -12,6 +12,1187 @@
...
+          </p>
+        </answer>
+      </task>
+    </activity>
+
+    <remark>
+      In <xref ref="deriving-pythagorean-identity-3"/>, we derived the third Pythagorean Identity: <m>\csc^2\theta-\cot^2\theta=1</m>. Note that we can rewrite what we got in the previous activity <m> \left( \cot^2\theta + 1 = \csc^2\theta \right)</m> to get to this form. 
+    </remark> 
+
+    <definition xml:id="def-pythagorean-identities">
```
siwelwerd, line 236: I think this (and others below) are really theorems rather than definitions.

source/precalculus/source/08-TE/01.ptx
```diff
@@ -12,6 +12,1187 @@
...
+          <p>
+            <m>30</m><degree/>
+          </p>
+        </answer>
+      </task>
+
+      <task>
+        <statement>
+          <p>
+            Substitute <m>\theta</m> with <m>30</m><degree/> in the half-angle formula for sine.
```
siwelwerd, line 1147: The variable should match what is in the formula

source/precalculus/source/08-TE/01.ptx
```diff
@@ -12,6 +12,1187 @@
...
+    </activity>
+
+    <remark>
+      In <xref ref="derive-difference-formula-for-cosine"/>, we derived the difference formula for cosine. The development of the sum/difference formulas are similar for sine and cosine and same for <m>\tan(\alpha-\beta)</m> compared to <m>\tan(\alpha+\beta)</m>. 
+    </remark>
+
+    <definition xml:id="def-sum-and-difference-formulas">
+      <statement>
+        <p>
+          The <term>Sum and Difference Formulas</term> in trigonometry are used to find the value of the trigonometric functions at specific angles where it is easier to express the angle as the sum or difference of unique angles such as <m> 0 </m> <degree/>, <m>30</m><degree/>, <m>45</m><degree/>, <m>60</m><degree/>, <m>90</m><degree/>, and <m>180</m><degree/>.
```
siwelwerd, line 520:
```suggestion
          The <term>Sum and Difference Formulas</term> in trigonometry are used to find the value of the trigonometric functions at specific angles where it is easier to express the angle as the sum or difference of unique angles such as <m> 0^\circ </m>, <m>30^\circ</m>, <m>45^\circ</m>, <m>60^\circ</m>, <m>90^\circ</m>, and <m>180^\circ</m>.
```

siwelwerd (changes requested):
See above comments!  Overall looks good though.

source/precalculus/source/08-TE/01.ptx
siwelwerd: Note that `<degree/>` [is intended]([Math as Text](https://pretextbook.org/doc/guide/html/topic-mathematics-practices.html#topic-mathematics-practices-9) 
https://pretextbook.org/doc/guide/html/topic-mathematics-practices.html#topic-mathematics-practices-9) for non-mathematical uses.  Since you are typesetting the numbers as math, just use `\circ` in the math.  This prevents line breaks between the number and the degree symbol.

source/precalculus/source/08-TE/01.ptx
tdegeorge: According to our outline, we wanted to cover the first identity in TR5 (which has not been written yet). I didn't want to link yet until it was written.

source/precalculus/source/08-TE/01.ptx
tdegeorge: How do I make it so that the text is not all italic? I want the "note" part to be regular text.

source/precalculus/source/08-TE/01.ptx
```diff
@@ -12,6 +12,1187 @@
...
+    </remark>
+
+    <definition xml:id="def-half-angle-formulas">
+      <statement>
+        <p>
+          The <term>Half Angle Formulas</term> are:
+          <ul>
+            <li>
+              <p>
+                <m>\sin\frac{\alpha}{2}=\pm\sqrt{\frac{1-\cos\alpha}{2}}</m>
```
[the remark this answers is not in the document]
tdegeorge: I don't disagree but I feel like most other texts list the formulas with both. Is it better with the note?

tdegeorge:
I believe I addressed all comments....double check and let me know!

source/precalculus/source/08-TE/01.ptx
StevenClontz: The styling of notes vs theorems vs definitions, etc, is not an author toggle and cannot be done for just one instance. It's best to choose a tag that matches is the semantic purpose of the block, and if you don't like the styling we need to discuss changing it globally across the books.

source/precalculus/source/08-TE/01.ptx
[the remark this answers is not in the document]
siwelwerd: It's better, but it's still not correct.  It suggests there are two values for a particular theta.  Compare this to the way we use $\pm$ in the quadratic formula, where it is important that there really are two values there.

Also, just because other texts do something does not bind us to do it that way.  We can be better and not write incorrect things.

siwelwerd:
> I believe I addressed all comments....double check and let me know!

If this is potentially ready to be merged (pending the double check), please mark as "Ready for Review" (and I'll look at it tomorrow).  This is still labeled as "Draft", which carries a presumption that you might still be working on it.

source/precalculus/source/08-TE/01.ptx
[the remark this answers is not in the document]
tdegeorge: I believe I fixed the issues we had on this but it's been a while since we talked about it. Can you double check what I did is good now?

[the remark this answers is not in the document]
siwelwerd: I think [Theorem 8.1.26](https://3e56e8a3.tbil.pages.dev/preview/precalculus/instructor/TE1#thm-half-angle-formulas) still needs to be updated to match what is in the [appendix]https://tbil.org/preview/precalculus/instructor/list-of-trig-identities.html#identity-half-angle)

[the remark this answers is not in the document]
siwelwerd: It could perhaps also do with a fluency builder at the end that makes use of one of the negative versions (I feel like we discussed that at some point but not sure it got documented or implemented anywhere).

siwelwerd (approved):

PR #401, opened 2024-11-05
Create a codechat config file for new codespace

StevenClontz:


siwelwerd (approved):

siwelwerd:
I see this is left as draft so I did not merge, but I fired up a fresh codespace and it seemed to work as I imagine you intended so I approved.

PR #410, opened 2024-11-11
Small edit to EL1

AbbyANoble:


source/precalculus/source/05-EL/01.ptx
```diff
@@ -294,7 +294,7 @@
           <me>
             f(x)=ab^{x}
```
jbunn3, line 295: Do you want a comma after the equation?

AbbyANoble: I could go either way, but it probably does need one.

tdegeorge: I also don't have a preference....we should probably look at how we have been doing it in other sections to be consistent.

tdegeorge (approved):

siwelwerd:
@tdegeorge Now that you've approved this, you'll want to hit the big green "Squash and Merge" button on this page to actually merge it into the main branch (and that will also close this PR)

PR #416, opened 2024-11-12
Appendix of common graphs

### LINKED ISSUE #414: Appendix idea
A page with graphs of parent functions (organized by type) would be a helpful reference. (And once it's made, link it to PF1 where I'm referencing it.) Writing this quickly, so I could be missing some, but I'm thinking:

Polynomial
$f(x)=x$
$f(x)=x^2$
$f(x)=x^3$

Other 
$f(x)=|x|$
$f(x)=\sqrt{x}$

Rational
$f(x)=\dfrac{1}{x}$

Exponential and logarithmic
$f(x)=b^x$
$f(x)=\log_b x$
$f(x)=e^x$
$f(x)=\ln x$

Trig
all six trig functions
all six inverse trig functions

siwelwerd:
Closes #414 

Suggest we get the first graph looking like we like before adding more.

@StevenClontz Is there a good way to put a title (or title-looking thing) on the figure, so that the equation is displayed above the graph and not just in the caption?


source/precalculus/source/meta/graphs.ptx
```diff
@@ -0,0 +1,17 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<appendix xml:id="graphs-of-common-functions" xmlns:xi="http://www.w3.org/2001/XInclude">
+    <title>Graphs of Common Funtions</title>
```
AbbyANoble, line 3:
```suggestion
    <title>Graphs of Common Functions</title>
```

source/precalculus/source/meta/graphs.ptx
```diff
@@ -0,0 +1,17 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<appendix xml:id="graphs-of-common-functions" xmlns:xi="http://www.w3.org/2001/XInclude">
+    <title>Graphs of Common Functions</title>
+    <paragraphs>
+        <title>Polynomial Functions</title>
+            <figure>
+                <image>
+                    <sageplot>
+                        f=x
+                        plot(f,xmin=-5,xmax=5,ymin=-5,ymax=5,thickness=3,aspect_ratio=1,gridlines=True, ticks=[[-5..5],[-5..5]])
```
StevenClontz, line 10:
```suggestion
f=x
p = plot(f,xmin=-5,xmax=5,ymin=-5,ymax=5,thickness=3,aspect_ratio=1,gridlines=True, ticks=[[-5..5],[-5..5]])
p += text(r'$f(x)=x$', (-3.5, 3.5), fontsize="x-large")
p
```

Only way I can think of to guarantee to have the function appear in a particular place

StevenClontz:
> @StevenClontz Is there a good way to put a title (or title-looking thing) on the figure, so that the equation is displayed above the graph and not just in the caption?

My most clever suggestion is to add it to the image itself using https://doc.sagemath.org/html/en/reference/plotting/sage/plot/text.html though I think https://9b40d30b.tbil.pages.dev/preview/precalculus/instructor/graphs-of-common-functions looks fine as-is.

StevenClontz (approved):

tdegeorge (approved):
I think it looks good. We might want to add the cube root function too.

jbunn3:
These turned out really nice!

PR #426, opened 2024-11-14
refactor EL2 to produce nice TeX

### LINKED ISSUE #409: EL2 CheckIt
Assigning Drew to begin with since he fixed previous issues that were similar, but feel free to unassign/reassign.

https://tbil.org/preview/precalculus/exercises/#/bank/EL2/1/

**Task 1:** 
When the base is a fraction, like versions 4 and 5, use the notation $\left(\dfrac{1}{6}\right)^x$ rather than $\dfrac{1}{6^x}$ to match the notation in the book.

**Task 3**
Similar to #374, #340. Fix the graphs so the windows are more standard looking and include the origin.

StevenClontz:
Closes #409

### DIFF
--- a/source/precalculus/exercises/outcomes/EL/EL2/generator.sage
+++ b/source/precalculus/exercises/outcomes/EL/EL2/generator.sage
@@ -1,41 +1,59 @@
 class Generator(BaseGenerator):
   def data(self):
-    a=1
-    b = choice([2,3,4,5,6])
-    c=0
-    d=0
-    k=1
-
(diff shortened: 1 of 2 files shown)


siwelwerd:
I just clicked through them all and it looks good to me.  

image

I noticed they look a little crowded by the bounding box -- maybe that's a Checkit thing though?

siwelwerd (approved):

StevenClontz:
Yeah that's a checkit app issue; nothing to be done on this end. (Should look fine in LMS and LaTeX exports.)

PR #428, opened 2024-11-16
AT3: Add common aliases for kernel and image

StevenClontz:


source/linear-algebra/source/03-AT/03.ptx
```diff
@@ -243,7 +244,9 @@ Which of these subspaces of <m>\IR^3</m> describes the set of all vectors that a
     <statement>
         <p>
 Let <m>T: V \rightarrow W</m> be a linear transformation.
-The <term>image</term> of <m>T</m> is an important subspace of <m>W</m> defined by
+The <term>image</term> of <m>T</m>
+(sometimes known as the <term>column space</term> of <m>T</m>)
```
StevenClontz, line 248:
```suggestion
(also known as the <term>column space</term> of <m>T</m>)
```

jbunn3:
I think it is really helpful to insert these aliases.

jkostiuk (changes requested):
I like the idea of introducing these aliases, but referring to the image of T as the columns space doesn't make sense at this point narratively. 

My suggestion is to keep the language related to the image as it is currently published, but then to add a comment to Observation 3.3.11 and/or Fact 3.3.12 introducing this alias. 

Alternatively (or better, in addition to this) I might also suggest that we simply define the column space of a matrix earlier in the book (I think EV6 seems an appropriate spot). This way, Fact 3.3.12s second bullet can simply state "The image of T is equal to column space of A".

jford1906:
I'm with Jordan on this one. I think we can keep it with these changes, but define the column space earlier in the text.

source/linear-algebra/source/03-AT/03.ptx
```diff
@@ -243,7 +244,9 @@ Which of these subspaces of <m>\IR^3</m> describes the set of all vectors that a
     <statement>
         <p>
 Let <m>T: V \rightarrow W</m> be a linear transformation.
-The <term>image</term> of <m>T</m> is an important subspace of <m>W</m> defined by
+The <term>image</term> of <m>T</m>
+(also known as the <term>column space</term> of <m>T</m>)
```
jkostiuk, line 248:
```suggestion
(also known as the <term>colu space</term> of <m>T</m>)
```

jkostiuk, line 248:
```suggestion
```

jford1906: I don't see this change in the preview, but I'm ok with it. I would also like to put an index tag here, since the index only takes it to the later section. I'm not sure if I can do that on an existing pull request?

StevenClontz:
@jford1906 has volunteered to adopt this to add in "column space" appropriately.

StevenClontz:
Oops I re-requested @jkostiuk review after I saw @jford1906 marked this as ready for review, but I don't see any changes? Did you forget to push?

jford1906:
I thought I did it right. I'll be home in about 90 minutes and can check.

jford1906:
Ok, see if the changes show up now

source/linear-algebra/source/03-AT/03.ptx
```diff
@@ -243,10 +244,13 @@ Which of these subspaces of <m>\IR^3</m> describes the set of all vectors that a
     <statement>
         <p>
 Let <m>T: V \rightarrow W</m> be a linear transformation.
-The <term>image</term> of <m>T</m> is an important subspace of <m>W</m> defined by
+The <term>image</term><idx>image</idx> of <m>T</m>
+is an important subspace of <m>W</m> defined by
 <me>
 \Im T = \left\{ \vec{w} \in W\ \big|\ \text{there is some }\vec v\in V \text{ with } T(\vec{v})=\vec{w}\right\}
 </me>
+Note that, if the transformation <m>T</m> is represented by the matrix <m>A</m>,
```
jkostiuk, line 252: I don't like that this statement spoils activity 3.3.10. Can we move this into the observation or fact that follows 3.3.10?

jkostiuk (changes requested):
otherwise I like the changes

source/linear-algebra/source/03-AT/03.ptx
jford1906: I moved this to the end of Observation 3.3.11

jford1906 (approved):

StevenClontz:
I believe this is ready to merge once @jkostiuk approves.

jkostiuk (approved):

PR #438, opened 2024-12-02
Fix latex presentation of log_b in EL4 checkit

### LINKED ISSUE #437: EL4 Check It
The logarithmic equations in EL4 are nonstandard in their display. They include things like a adding a negative number (without parentheses) instead of subtraction. And +0 at the end of some. My guess is that the equation was defined in pieces, and the pieces were inserted in an outline that had a constant at the end.

The ones below are versions 1 and 2 from https://tbil.org/preview/precalculus/exercises/#/bank/EL4/1/.  

Screenshot 2024-12-01 at 4 22 18 PM

Screenshot 2024-12-01 at 4 23 38 PM

I'd prefer them to look more standard, like $-\log_2 (x) -3$ and  $-\log_4 (x+2)$.

siwelwerd:
Closes #437

Not 100% sure this is the best way to do this, but it seems to work here.


siwelwerd:
@AbbyANoble Check out the slick preview above that will make this easy to review!

source/precalculus/exercises/outcomes/EL/EL4/generator.sage
```diff
@@ -22,10 +23,8 @@ class Generator(BaseGenerator):
       d = choice(range(1,6))*choice([-1,1])
     
     f=a*log(k*x-c)/log(b)+d
-    if a == 1:
-      f_string = f"\\log _{{ {b} }} \\left({k*x-c}\\right)+{d}"
-    else:
-      f_string = f"-\\log _{{ {b} }} \\left({k*x-c}\\right)+{d}"
+    logb=function_factory('logb',1,'\\log _{'+str(b)+'}')
```
StevenClontz, line 26:
```suggestion
    logb = function_factory("logb", 1, f"\\log _{{ {b} }}")
```

Nice find btw

siwelwerd: This is cleaner, thanks.

AbbyANoble:
> @AbbyANoble Check out the slick preview above that will make this easy to review!

THANKS! I could never remember how to preview a CheckIt, so I would always put those off on future Abby. 🤣

AbbyANoble (approved):

PR #439, opened 2024-12-03
Add sage images for AD

cg2wilson:
Sage images for AD added!


source/calculus/source/03-AD/08.ptx
```diff
@@ -29,18 +29,252 @@
...
+                    geometry += sll2
+                    geometry += label_side
+
+                    # Set axes and title
+                    geometry.set_axes_range(-1, side_large_square + 1, -1, side_large_square + 1)
+                    geometry.axes(False)
+
+                    # Show the plot
+                    geometry
+                  </sageplot>
```
StevenClontz, line 90: Old:

image

New:

image

I think we should increase the font size in the new one a bit

source/calculus/source/03-AD/08.ptx
```diff
@@ -199,7 +433,30 @@
...
+              geometry += front
+              geometry += front_border
+              geometry += back
+              geometry += back_border
+              geometry += right
+              geometry += right_border
+              geometry.axes(False)
+              geometry
+            </sageplot>
+          </image>
```
StevenClontz, line 459: Old:

image

New:

image

Maybe it's just me, but I think its easier to understand the image when the darker side is on bottom.

StevenClontz: If possible, using dashes for sides obscured to the viewer might help as well.

source/calculus/source/03-AD/08.ptx
cg2wilson: i increased the font size to 15pt- i can go larger if people think it's necessary.

source/calculus/source/03-AD/08.ptx
cg2wilson: the most recent commit changes the shading (and doesn't remove the helper axes, oops), and i'm working on adding the dashed lines

StevenClontz (approved):

PR #450, opened 2024-12-16
Scenario instead of settings

StevenClontz:
More information on why this is necessary...

### DIFF
--- a/source/calculus/source/03-AD/04.ptx
+++ b/source/calculus/source/03-AD/04.ptx
@@ -13,7 +13,7 @@
     <title>Activities</title>
     <remark>
     <p>
-      In many different scenarios,
+      In many different situations,
       we are interested in knowing where a function achieves its least and greatest values.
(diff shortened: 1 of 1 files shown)


source/calculus/source/03-AD/04.ptx
```diff
@@ -13,7 +13,7 @@
     <title>Activities</title>
     <remark>
     <p>
-      In many different settings,
+      In many different scenarios,
```
siwelwerd, line 16:
```suggestion
      In many different situations,
```

What about "situations"?

source/calculus/source/03-AD/04.ptx
```diff
@@ -13,7 +13,7 @@
     <title>Activities</title>
```
StevenClontz, line 13:
```suggestion
    <title>Puzzles</title>
```

StevenClontz: eh never mind

siwelwerd:
This was for training purposes only, closing to declutter

Issue #334, opened 2024-10-02
Suggestion for LE2


jkostiuk:
Based on a conversation with Drew: add a fourth activity to the sequence 1.2.12,13,14 that:

- gives the students the matrix in 1.2.12 (unknowingly)
- asks them to mark the pivot position
- ask them which row operation they'd use and why

siwelwerd:
We were discussing that this is basically the conversation we like to have with them as a wrap-up discussion, but instead of us telling them why not have them do it as an activity?

jkostiuk:
adding on while it's in my brain: I think we should consider re-structuring this in PreTeXT as a single activity with 4 parts. 

Personally, I think an "Activity" should have a satisfying resolution. When I facilitate this, the resolution doesn't come until the sequence is complete and so the whole sequence feels like one activity as far as I'm concerned.

PR #460, opened 2024-12-16
Suggestion for LE2 #334

jkostiuk:
I reorganized the activities beginning with 1.2.12 into a single activity with multiple parts.

For facilitators, I think this is helpful as it reinforces that these four tasks are part of one unified thread of inquiry. It now provides, explicitly, an opportunity for Facilitators to have students revisit the first matrix again after unpacking student reasoning in the first three tasks.

### DIFF
--- a/source/linear-algebra/source/01-LE/02.ptx
+++ b/source/linear-algebra/source/01-LE/02.ptx
@@ -721,6 +721,13 @@ in many applications we use it for in <xref ref="EV"/>.
         </ol>
     </task>
 </activity>
+<commentary component="instructor">
+<paragraphs>
+   <title>Instructor Note</title>
+   <p> The fact that the matrix considered in part (d) is the same as in (a) is intentional. 
+   As you're facilitating this discussion with your students, explore how your students' reasoning has changed in coming to an answer in (d) compared to how they answered in (a).</p>
+</paragraphs>
+</commentary>
 
 <observation>
     <p>
--- a/source/linear-algebra/source/01-LE/02.ptx
+++ b/source/linear-algebra/source/01-LE/02.ptx
@@ -646,7 +646,21 @@ in many applications we use it for in <xref ref="EV"/>.
 
 
 
-
+<commentary component="instructor">
+<paragraphs>
+   <title>Instructor Note</title>
+    <p> 
+        In the following 4-part activity, students make decisions about which row operations are most helpful for them in producing the RREF. 
+        The matrices appearing in parts (b) and (c) are obtained by applying specific choices of row operations to the matrix appearing previously; the matrix in part (d) is the one that we started with. 
+    </p>
+    <p>
+        Our experience in facilitating these activities is that student reasoning in part (a) is largely grounded in the production of zeroes, no matter how they appear; because of this, all of the choices in part (a) may appear to be equally valid to them. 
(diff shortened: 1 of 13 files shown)


source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -644,64 +644,82 @@ in many applications we use it for in <xref ref="EV"/>.
...
+            <li> Add row 3 to row 2 (<m>R_2+R_3 \rightarrow R_2</m>)
+            </li>
+            <li> Add row 2 to row 3 (<m>R_3+R_2 \rightarrow R_3</m>)
+            </li>
+            <li> Swap row 1 to row 2 (<m>R_1 \leftrightarrow R_2</m>)
+            </li>
+            <li> Add -2 row 2 to row 1 (<m>R_1-2R_2 \rightarrow R_1</m>)
+            </li>
+        </ol>
+    </task>
```
StevenClontz, line 722: This appears to be accidentally identical to the first task.

StevenClontz: oh wait well maybe I should read your PR description. Then again, instructors won't read the PR description either - I don't see how this repetition helps just by reading the activity book, but it's late and I should look again with fresh eyes in the morning.

jkostiuk: @siwelwerd can you clarify what you're proposing. I did a few things and I'm not clear if we're rejecting all of them. 

First I grouped the activities into one multi-part activity (do we like this or not?)

Secondly, are you suggesting adding a separate activity in which we revisit the first activity and then include an instructor note?

siwelwerd: I'm saying to reject what Steven proposed, but add an instructor note as to why it is repeated, as some instructors might think it's an error

jkostiuk: working on the note now.

siwelwerd:
@jkostiuk I updated the style guide with the proper code for an instructor note

StevenClontz:
It's still not clear to me what we're guiding students to inquire into by repeating the first task. What is the change you expect the students to make before/after?

jkostiuk:
In my teams experience with this activity, students answers are all over the place the first time they see the matrix. Some want the 1 in the top left, but a bunch of others choose other options because they make a lot of zeroes. 

In the next two activities, students actually use the marked pivots strategically to clear the positions above and below. 

So, by revisiting, I’m hoping students might actually use the pivot as part of the strategy when they weren’t usually doing so before. 

Also, as an aside, I don’t actually expect students to recognize that it’s the same matrix.

siwelwerd:
Maybe expanding the instructor note, and moving it to before the activity, would help.  With the callback, it's probably okay to leave the first task a little unresolved, and then resolve it with part (d).  But it would be helpful to write that out for instructors.

jkostiuk:
took a stab at an expanded note @siwelwerd

siwelwerd:
I think that's good!  @StevenClontz, is that more clear to you what is happening now?  Jordan and I talked about this previously so my view is colored by those conversations.

StevenClontz:
Okay, with the expanded note, I now get the rationale for the revisit.

I think I have an idea for an alternative here we might consider, but that should not block this strict improvement.

StevenClontz (approved):

PR #462, opened 2024-12-16
Add scaffolding and reorder exercises in TI1

daverosoff:
This addresses #54.


StevenClontz (commented):
New activity: [Activity 5.1.13](https://177cbb93.tbil.pages.dev/preview/calculus/instructor/TI1#TI1-3-14)

Existing activity: [Activity 5.1.9](https://library.tbil.org/2024/calculus/TI1.html#TI1-3-10)

I agree with @AbbyANoble's note in #54 that the side-by-side squeezes things a bit together, and that moving the activity to later in the section might be better, after students have seen some more examples.

However, I'm unsure I like all the commentary being added here. When I wrote the activity, I wanted students to see two clean, correct writeups to help them build fluency in how to work though a substitution problem, and it's a mix of instructor facilitation and student inquiry to break down each step of the two writeups.

I'd propose making task (a) the first writeup (the left panel of the original sidebyside), with an MCQ asking if it was a valid solution or not. Then task (b) asks the same of the second writeup (the right sidebyside panel). Then task (c) asks students to justify why both solutions are correct.

daverosoff:
OK, is there a way I should have known to ignore the part of #54 about "more scaffolding" but pay attention to the rest? Thanks.

StevenClontz:
Not without having a discussion. :-/ Speaking of which, @AbbyANoble should chime in with her opinion.

jbunn3:
I looked over the suggested changes here and discussion, and I think having this a bit later is a good call. I'm not so sure I think this problem would need scaffolding at this point of the section, as by the time it surfaces, I would expect the students to have gained some experience in choosing the substitution. However, it did appear to be a bit too early in the previous version. 

I would suggest keeping the new ordering, but itemizing the two original, clean write-up so that they appear one after the other. Then, I would keep part (c) where they analyze the different representations of the solution. 

(To be honest, I'm not a big fan of having the u-sub work in a second column, although that's how many people do it.)

StevenClontz (changes requested):
See above recommended changes.

AbbyANoble:
I think the scaffolding is helpful here because, even though it's moved down, it's still the first time they see a "weird" choice for u. It's really the only time in the whole sections where the u is somewhat unexpected. @StevenClontz I get what you mean about leaving it like you had it to help with fluency, but if there is no example of this happening before (with at least some scaffolding), I think the fluency part is very hard to catch on to by just looking at it for an average calc student at this point. @daverosoff's adds in the thought process behind it, which I like. 

I'd move this to the very end of the section, though, then add a couple fluency problems that have this similar idea of unexpected substitutions.

Then I'd move [Activity 5.1.21](https://177cbb93.tbil.pages.dev/preview/calculus/instructor/TI1#TI1-3-22) (shown below) after that, probably with making the variable something other than u. (I know it doesn't really matter, but trying a new type where the variable is the one you usually use to substitute seems like adding in unnecessary complexity.) The way I did this one I chose the "expected" u or x^3+1, but then you have to fiddle around to substitute everything else in a way that hasn't happened before in the activities. (Unless I'm missing something.)

Screenshot 2024-12-18 at 11 08 13 AM

Then some fluency with these as well. Or maybe put this example right after @daverosoff's and do one fluency set of the unexpected substitutions.

daverosoff:
Looks like I didn't actually finish this. I feel silly. Anyway, this branch is too old to be convenient—I'll start a fresh one and work on the changes requested above.

PR #470, opened 2024-12-17
Fixed One-to-one property of logs in EL3, EL5, and EL6 #430

### LINKED ISSUE #430: One-to-one property of logs in EL3, EL5, and EL6
We define this property in [Definition 5.6.7](https://library.tbil.org/2024/precalculus/EL6.html#def-one-to-one-property), and I think it would help to write it out in the notation instead of just words: 

$log_b M = log_b N$ if and only if $M=N$.

Also, I would suggest changing the second half of the definition into an Observation. Something like "Notice this definition allows us to set the arguments (M and N) equal to each other once we get each side expressed as a single logarithm with matching bases."

Also, this seems to be the first time it's defined (in EL6), but it is mentioned in the previous section (EL5) as if it was defined in EL3. See

- [Remark 5.5.1](https://library.tbil.org/2024/precalculus/EL5.html#EL5-3-3) 
- [Activity 5.5.2(b)](https://library.tbil.org/2024/precalculus/EL5.html#investigate-product-property-of-logs-3) 
- [Activity 5.5.3(b)](https://library.tbil.org/2024/precalculus/EL5.html#investigate-quotient-property-of-logs-3) 

We need to clear this up and figure out where we want everything located and referenced.

tdegeorge:
Fixed #430


source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,98 @@
...
+      </statement>
+      <answer>
+        <p>
+          Based on what we have seen in parts (b) and (c), we can conclude that <m>M=N</m>.
+        </p>
+      </answer>
+    </task>
+    </activity>
+
+    <definition xml:id="def-one-to-one-property">
```
AbbyANoble, line 86: Do we have a tag for property to use here instead of definition?

siwelwerd: Nope.  Maybe `<fact>` will work?

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,98 @@
...
+      <statement>
+        <p>
+          The <term>one-to-one property of logarithms</term> states that if both sides of an equation can be rewritten as a single logarithm with the same base, then the arguments can be set equal to each other.
```
AbbyANoble, line 89: The paragraph here is an if-then statement, but then the part below is an if and only if, so they are sort of saying different things.  I'd suggest changing the paragraph to just set up the parameters for the variables of the if and only if, and then the observation below is what the paragraph said.

```suggestion
         For any values <m>S>0</m> and <m>T>0</m>,  and where <m>b>0</m> and <m>b\neq 1</m>, 
         
         <p>
            <me>\log_bM=\log_bN</me> if and only if <me>M=N.</me>
          </p>
         
        <p> This is called the <term>one-to-one property of logarithms</term></p>
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,98 @@
 <introduction>
   In this section, we will explore the properties of logarithms and learn how to manipulate them that will be helpful when we are ready to solve logarithmic equations.
   </introduction>
+
+  <remark>
+      Recall from <xref ref="EL3"/> that we can convert between exponential and logarithmic forms. <me>\log_bx=y</me> is equivalent to <me>b^y=x</me>
```
AbbyANoble, line 19:
```suggestion
     <p> Recall from <xref ref="EL3"/> that we can convert between exponential and logarithmic forms. <me>\log_bx=y</me> is equivalent to <me>b^y=x</me>.</p>
```

StevenClontz:
This pull request was not created correctly - it will merge into a feature branch rather than `main` 😭:

image

StevenClontz:
I think this should work now. @AbbyANoble needs to leave an approving review before we can merge.

source/precalculus/source/05-EL/03.ptx
```diff
@@ -200,14 +203,19 @@
...
+        <p>
+          Range: <m>(-\infty, \infty)</m>
         </p>
       </answer>
   </task>
 </activity>
+
 <remark>
-  The powers of 10 function <m>P(t)</m> has an inverse <m>L</m>.  This new function <m>L</m> is called the base 10 logarithm.  But, we could have done a similar procedure with any base, which leads to the following definition.
+  The powers of <m>10</m> function <m>P(t)</m> has an inverse <m>L</m>.  This new function <m>L</m> is called the base <m>10</m> logarithm.  But, we could have done a similar procedure with any base, which leads to the following definition.
```
AbbyANoble, line 216:
```suggestion
<p>  The powers of <m>10</m> function <m>P(t)</m> has an inverse <m>L</m>.  This new function <m>L</m> is called the base <m>10</m> logarithm.  But, we could have done a similar procedure with any base, which leads to the following definition. </p>
```

source/precalculus/source/05-EL/03.ptx
```diff
@@ -219,11 +227,11 @@
   </statement>
 </definition>
 <remark>
```
AbbyANoble, line 229:
```suggestion
<remark xml:id="remark-log-exp-form">
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,104 @@
 <introduction>
   In this section, we will explore the properties of logarithms and learn how to manipulate them that will be helpful when we are ready to solve logarithmic equations.
   </introduction>
+
+  <remark>
+     <p> Recall from <xref ref="EL3"/> that we can convert between exponential and logarithmic forms. <me>\log_bx=y</me> is equivalent to <me>b^y=x</me>.</p>
```
AbbyANoble, line 19:
```suggestion
     <p> Recall from <xref ref="remark-log-exp-form"/> that we can convert between exponential and logarithmic forms. <me>\log_bx=y</me> is equivalent to <me>b^y=x</me>.</p>
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,104 @@
...
+      </statement>
+      <answer>
+        <p>
+          Based on what we have seen in parts (b) and (c), we can conclude that <m>M=N</m>.
+        </p>
+      </answer>
+    </task>
+    </activity>
+
+    <definition xml:id="def-one-to-one-property">
```
AbbyANoble, line 86:
```suggestion
    <fact xml:id="def-one-to-one-property">
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,104 @@
...
+            <me>\log_bM=\log_bN</me> if and only if <me>M=N.</me>
+          </p>
+         
+        <p> This is called the <term>one-to-one property of logarithms</term></p>
+          <p>
+            <me>\log_bM=\log_bN</me> if and only if <me>M=N</me>
+          </p>
+        </p>
+      </statement>
+    </definition>
```
AbbyANoble, line 101:
```suggestion
    </fact>
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,104 @@
...
+
+    <observation>
+      <p>
+        Notice this definition allows us to set the arguments (<m>M</m> and <m>N</m>) equal to each other once we get each side expressed as a single logarithm with matching bases.
```
AbbyANoble, line 105:
```suggestion
        Notice this fact will eventually help us solve logarithmic equations. If we have an equation where each side is expressed as a single logarithm with matching bases (such as <m>\log_b M = \log_bN</m>), then it follows that the arguments (<m>M</m> and <m>N</m>) are also equal to each other.
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,104 @@
...
+      </p>
+    </observation>
    
   <remark>
-    Recall that <m>\log_b M=\log_b N</m> if and only if <m>M=N</m>. In addition, because exponentials and logarithms are inverses, we also know that <m>\log_b(b^{k})=k</m>.
-    In addition, according to the law of exponents, we know that: <me>x^a \cdot x^b=x^{a+b}</me>
+    Recall that exponentials and logarithms are inverses. We know that <m>\log_b(b^{k})=k</m> and according to the law of exponents, we know that: <me>x^a \cdot x^b=x^{a+b}</me>
```
AbbyANoble, line 110:
```suggestion
 <p>   Recall that exponential functions and logarithmic functions are inverses. We know that <m>\log_b(b^{k})=k</m> and according to the law of exponents, we know that: 
 <me>x^a \cdot x^b=x^{a+b}</me>
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,104 @@
...
+      </answer>
+    </task>
+    </activity>
+
+    <definition xml:id="def-one-to-one-property">
+      <statement>
+        <p>
+         For any values <m>S>0</m> and <m>T>0</m>,  and where <m>b>0</m> and <m>b\neq 1</m>, 
+         
+         <p>
```
AbbyANoble, line 91:
```suggestion
         
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,104 @@
...
+            <me>\log_bM=\log_bN</me> if and only if <me>M=N.</me>
+          </p>
+         
+        <p> This is called the <term>one-to-one property of logarithms</term></p>
```
AbbyANoble, line 95:
```suggestion
        <p> This is called the <term>one-to-one property of logarithms</term>.</p>
```

source/precalculus/source/05-EL/05.ptx
```diff
@@ -14,14 +14,104 @@
...
+          <p>
+            <me>\log_bM=\log_bN</me> if and only if <me>M=N</me>
+          </p>
+        </p>
```
AbbyANoble, line 99:
```suggestion

```

source/precalculus/source/02-FN/05.ptx
AbbyANoble, line 1: I'm not sure why this file is in this pull request? @siwelwerd or @StevenClontz?

siwelwerd: I think Tonya accidentally branched off of a branch, but I think it's fixed now

siwelwerd:
I merged all of your commits @AbbyANoble , but I think it's still a little hard to follow since the remarks aren't showing up right.  Maybe @StevenClontz  can merge #514 into main to fix those, and then merge main into this so it's easier to see.

StevenClontz:
Try that...

siwelwerd:
@AbbyANoble Please review latest preview build, and see if there is any remaining comment from #430 discussion that needs to be addressed here.

siwelwerd:
@AbbyANoble once this preview builds, it should show all of the work with this branch along with Jordan's revised remark from #523

siwelwerd:
> This part of Jordan's revision doesn't show up in this preview.

Okay it's back there now.  I must have hit a wrong button.

PR #479, opened 2024-12-17
Bunn-12.17.24-formatting-in-octave

jbunn3:
I wanted to find a later place in the text where it made sense to introduce this. Suggestions are welcome.

### DIFF
--- a/source/linear-algebra/source/02-EV/05.ptx
+++ b/source/linear-algebra/source/02-EV/05.ptx
@@ -241,8 +241,10 @@ In particular, the standard basis for <m>\mathbb R^3</m> is <m>\{\vec e_1,\vec e
   </definition>
 
 <commentary component="instructor">
-            <title>Instructor Note</title>
-                <p> Using format rat aligns the columns nicely and actually converts decimals to fractions, when applicable.</p>
+<paragraphs>
+   <title>Instructor Note</title>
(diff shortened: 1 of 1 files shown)


jbunn3:
The commentary code 
```
<commentary component="instructor">
  <title>Instructor Note</title>
  <p>Your note goes here.</p>
</commentary>
```
doesn't seem to be working.

siwelwerd:
You have it within an activity right now, which may not be allowed.  Try moving it outside the activity

jbunn3:
> You have it within an activity right now, which may not be allowed. Try moving it outside the activity

I was wondering if that was the culprit.

siwelwerd:
Actually looking again, it's within a `<p>` as well which is almost certainly bad

jbunn3:
> Actually looking again, it's within a `<p>` as well which is almost certainly bad

I fixed that, but all my efforts to move it around doesn't seem to fix it.

siwelwerd:
Style guide has been updated with code that should work for the instructor note

jbunn3:
@StevenClontz Can you take a look at this? I'm not sure why an old branch of mine got involved in this PR? Did I do something wrong?

StevenClontz:
This was accidentally branched off fof `bunn-11.8.24-fix-sq7`, not `main`, which is why an approval wasn't necessary before the Merge button was available. I think #490 will fix this.

PR #486, opened 2024-12-18
Refactor LE2 generation

### LINKED ISSUE #20: Better generation of matrices to RREF by hand in LE2
In https://github.com/TeamBasedInquiryLearning/linear-algebra/blob/main/exercises/outcomes/GT/GT2/generator.sage we wrote some logic to "undo" how a student might compute the determinant by hand, ensuring problems have similar difficulties.

I'd like to implement something similar for LE2: start with the "nice" RREF, then apply row operations to produce an interesting result that covers all the row op techniques we want to see from students.

StevenClontz:
Closes #20

I'd like to mark the pivots, and asked at https://ask.sagemath.org/question/80691/mark-pivot-positions-in-a-matrix/ if this is possible, but don't know how to today.

### DIFF
--- a/source/linear-algebra/exercises/outcomes/LE/LE2/generator.sage
+++ b/source/linear-algebra/exercises/outcomes/LE/LE2/generator.sage
@@ -2,46 +2,64 @@ load("../sage/common.sage")
 
 class Generator(BaseGenerator):
     def data(self):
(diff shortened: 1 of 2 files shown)


jkostiuk (approved):
generated 20 exercises and they looked good to me. i like this change.

PR #490, opened 2024-12-18
Fix #479 merge

StevenClontz:
This should merge the result of #479 into `main` instead of the feature branch.


jbunn3 (approved):
Looks great!

PR #491, opened 2024-12-18
Fix LT2 activity answer scaffolding

StevenClontz:


StevenClontz:
Just FYI - this more or less looks fine in the HTML build without the fix, but I think it might be breaking our PDF builds.

PR #496, opened 2024-12-18
Display powers of trig functions in usual way

### LINKED ISSUE #39: Textbook versus Checkit notation Inconsistency for sin^n(x)
In TI3, we find integrals of the form `\int \sin^4(x)\cos^3(x) \, dx` but over in CheckIt the notation used is `\int \sin (x)^4 \cos (x)^3 \, dx`, e.g. compare [Activity 5.3.3](https://teambasedinquirylearning.github.io/calculus/2023/TI3.html#activity-290) 
https://teambasedinquirylearning.github.io/calculus/2023/TI3.html#activity-290 to https://teambasedinquirylearning.github.io/calculus/2023/exercises/#/bank/TI3/3/ . My preference would be for the textbook version of the notation, but I have strong feelings that the two platforms should align all notation whenever possible.

siwelwerd:
Closes #39

### DIFF
--- a/source/calculus/exercises/outcomes/TI/TI3/generator.sage
+++ b/source/calculus/exercises/outcomes/TI/TI3/generator.sage
@@ -1,33 +1,44 @@
 #Functions to display powers of trig functions
-def print_cosp(self,*args): return f"\\cos ^{args[1]}({latex(args[0])})"
-cosp = function("cosp",nargs=2,print_latex_func=print_cosp)
-def print_sinp(self,*args): return f"\\sin ^{args[1]}({latex(args[0])})"
-sinp = function("sinp",nargs=2,print_latex_func=print_sinp)
+def print_cosp(self,*args): return f"\\cos ^{{{args[1]}}}({latex(args[0])})"
+def deriv_cosp(self,*args,**kwds): 
+    if args[1]==1:
+        return -1*sinp(args[0],1)*args[0].derivative(args[kwds['diff_param']])
+    else:
+        return args[1]*-1*cosp(args[0],args[1]-1)*sinp(args[0],1)*args[0].derivative(args[kwds['diff_param']])
+
+cosp = function("cosp",nargs=2,print_latex_func=print_cosp,derivative_func=deriv_cosp)
+def print_sinp(self,*args): return f"\\sin ^{{{args[1]}}}({latex(args[0])})"
+def deriv_sinp(self,*args,**kwds): 
+    if args[1]==1:
+        return cosp(args[0],1)*args[0].derivative(args[kwds['diff_param']])
+    else:
+        return args[1]*sinp(args[0],args[1]-1)*cosp(args[0],1)*args[0].derivative(args[kwds['diff_param']])
+
+sinp = function("sinp",nargs=2,print_latex_func=print_sinp, derivative_func=deriv_sinp)
 
 class Generator(BaseGenerator):
 
     def data(self):
-        # integral with odd power
+        # Task 1, integral with odd power
         x=var("x")
         
-        even=2*randint(1,6)
-        
-        odd=choice([5,7])
+        n=randint(1,6)
+        m=randint(2,3)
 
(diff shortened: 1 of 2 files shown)


siwelwerd:
I think this is working!
image

However, I had to do the integration by hand and hardcode it in, so it would be helpful if someone could check the answers on these (particularly task 1).  I spot checked a handful of them and they looked okay though.

source/calculus/exercises/outcomes/TI/TI3/generator.sage
```diff
@@ -1,32 +1,50 @@
...
-        # integral with even powers
+        # Task 2, integral with even powers
 
         a = randrange(2,6)
         m = randrange(2,5)
         n = randrange(2,5)
         k = 2^randrange(2,5)
-        g = k*cos(a*x)^(2*m)*sin(a*x)^(2*n)
+        g = k*cosp(a*x,2*m)*sinp(a*x,2*n)
         also_g = k/2^(m+n)*(1+cos(2*a*x))^m*(1-cos(2*a*x))^n
```
StevenClontz, line 48:
```suggestion
        also_g = k/2^(m+n)*(1+cos(2*a*x))^m*(1-cos(2*a*x))^n
        assert f.diff() == also_g
```

Or something like that to ensure no errors were made? Or something that's more clearly the same as `g`?

StevenClontz (changes requested):
Does my suggestion to verify equality with code make sense? A spot-check looks good.

It's off-topic, but while we're here, should task 2 say something like "...identities to rewrite {{int}} into an integral with smaller exponents..."?

source/calculus/exercises/outcomes/TI/TI3/generator.sage
siwelwerd: I don't think that will work, the way `cosp` is created sage really doesn't know that it's a power of `cos`.  We lose a lot of that kind of thing by creating our own function.

StevenClontz: Hmm, it might be quicker to demo than explain what I'm trying to say, one sec

source/calculus/exercises/outcomes/TI/TI3/generator.sage
```diff
@@ -1,32 +1,50 @@
...
+
+cosp = function("cosp",nargs=2,print_latex_func=print_cosp,derivative_func=deriv_cosp)
+def print_sinp(self,*args): return f"\\sin ^{{{args[1]}}}({latex(args[0])})"
+def deriv_sinp(self,*args,**kwds): 
+    if args[1]==1:
+        return cosp(args[0],1)*args[0].derivative(args[kwds['diff_param']])
+    else:
+        return args[1]*sinp(args[0],args[1]-1)*cosp(args[0],1)*args[0].derivative(args[kwds['diff_param']])
+
+sinp = function("sinp",nargs=2,print_latex_func=print_sinp, derivative_func=deriv_sinp)
```
StevenClontz, line 17: We should open an issue to move this into our common library

source/calculus/exercises/outcomes/TI/TI3/generator.sage
siwelwerd: Maybe there is a way to use `assuming` to do something.  Maybe that's a separate issue t hough to go back and add checks like these wherever we are doing something by hand (across exercises)

StevenClontz: It was not quicker, so I think this is okay to use as-is.

StevenClontz (approved):

StevenClontz:
> > It's off-topic, but while we're here, should task 2 say something like "...identities to rewrite {{int}} into an integral with smaller exponents..."?
> 
> I didn't love that prompt either, but was focusing on getting the powers to display properly. I'm happy for you to suggest alternative language.

I'll make a quick separate PR

PR #497, opened 2024-12-18
Creating PS 5 on Taylor's Theorem

jbunn3:
This is the beginning stages of creating the section P5, starting from a sample problem, a learning outcome, and some introduction.

### DIFF
--- a/source/calculus/source/09-PS/05.ptx
+++ b/source/calculus/source/09-PS/05.ptx
@@ -107,17 +107,17 @@
                     </li>
                     <li>
                         <p>
-                            It will be greater than both <m>R_3(4)</m> and <m>R_4(4)</m>.
+                            It will be greater than both <m>R_2(4)</m> and <m>R_3(4)</m>.
                         </p>
                     </li>
                     <li>
                         <p>
-                            It will be between <m>R_3(4)</m> and <m>R_4(4)</m>.
+                            It will be between <m>R_2(4)</m> and <m>R_3(4)</m>.
                         </p>
                     </li>
                     <li>
                         <p>
-                            It will be less than both <m>R_3(4)</m> and <m>R_4(4)</m>.
+                            It will be less than both <m>R_2(4)</m> and <m>R_3(4)</m>.
                         </p>
                     </li>
                 </ol>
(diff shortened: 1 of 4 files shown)


source/calculus/source/09-PS/05.ptx
```diff
@@ -0,0 +1,361 @@
...
+                    <li>
+                        <p>
+                            It will be less than both <m>R_3(4)</m> and <m>R_4(4)</m>.
+                        </p>
+                    </li>
+                </ol>
+            </statement>
+        </task>
+    </activity>
+
```
cg2wilson, line 127: I think the answer choices are meant to be $R_2(4)$ and $R_3(4)$

source/calculus/source/09-PS/05.ptx
```diff
@@ -0,0 +1,361 @@
...
+                </me>
+                If there exists <m>M_k</m> such that <m>|f^{(k+1)}(x)|\leq M_k</m> for all <m>x</m> in <m>I</m>, then the error in 
+                the approximation <m>f(x)\approx T_k(x)</m> has an upper bound:
+                <me>
+                    |R_k(x)|\leq \dfrac{M_k}{(k+1)!}|x-a|^{k+1}.
+                </me>               
+                
+            </p>
+        </statement>
+    </theorem>
```
cg2wilson, line 163: Should the centering term be $x-c$ to be consistent with what's written in Fact 9.5.4?

source/calculus/source/09-PS/05.ptx
```diff
@@ -0,0 +1,361 @@
...
+                        <p>
+                            All are increasing.
+                        </p>
+                    </li>
+                     <li>
+                        <p>
+                            <m>|f'(x)|</m> and <m>|f'''(x)|</m> are decreasing, while <m>|f''(x)|</m> and <m>|f''''(x)|</m> are decreasing.
+                        </p>
+                    </li>
+                </ol>
```
cg2wilson, line 217: I think the fourth derivative notation should be consistent; I think $f^{(4)}(x)$ looks better than $f^{''''}(x)$, but it doesn't matter which is used, I don't think.

cg2wilson (changes requested):
Most of what I saw are small formatting/consistency things. The substance of it all looks great to me!

source/calculus/source/09-PS/05.ptx
jbunn3: Thanks, I made the updates.

source/calculus/source/09-PS/05.ptx
jbunn3: Good point. It has been updated.

source/calculus/source/09-PS/05.ptx
jbunn3: I had thought about but just left it. I agree keeping it consistent is better. It's updated.

cg2wilson (approved):

PR #498, opened 2024-12-18
FN2 CheckIt attempt

### LINKED ISSUE #495: FN2 CheckIt tweak
See Task 2 of https://tbil.org/preview/precalculus/exercises/#/bank/FN2/2/. 

- I think we should randomize the order of the two parts of task 2. That way students need to read and understand the notation rather than knowing the pattern. 
- For part 2 (Find the x-values...), the answer states a fact, but it is not actually the answer to the question asked. It's asking for x-values, so I think the answer reflect that better instead of just telling us something that is true. I'd recommend  
    - $f(x)=-4$ when $x=1.5, -3$, or simply
    - $x=1.5, -3$ 

I'm going to try to do the second part, but I don't know how to make the randomization happen.

AbbyANoble:
I tried to make the second part model what I mention in #495, but I don't know if it is correct.

Closes #495

### DIFF
--- a/source/precalculus/exercises/outcomes/FN/FN2/template.xml
+++ b/source/precalculus/exercises/outcomes/FN/FN2/template.xml
@@ -44,7 +44,7 @@
         <!-- {{#findx}}-->
         <knowl>
             <content>
-                <p>For which <m>x</m>-values in the interval <m>{{interval}}</m> does <m>{{gname}}(x)={{result}}</m>?
+                <p>For which <m>x</m>-value(s) in the interval <m>{{interval}}</m> does <m>{{gname}}(x)={{result}}</m>?
                 </p>
             </content>
             <outtro>
--- a/source/precalculus/exercises/outcomes/FN/FN2/template.xml
+++ b/source/precalculus/exercises/outcomes/FN/FN2/template.xml
@@ -44,9 +44,11 @@
                 </p>
             </content>
             <outtro>
+                <p><m>{{gname}}(x)={{result}}</m> when 
                 <!--{{#values}}-->
-                <p><m>{{gname}}(x)={{result}}</m>when <m>x={{x}}</m></p>
+                <m>x={{x}}</m>,
                 <!--{{/values}}-->
+                </p>
             </outtro>
         </knowl>
     </knowl>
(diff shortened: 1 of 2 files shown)


source/precalculus/exercises/outcomes/FN/FN2/template.xml
```diff
@@ -45,7 +45,7 @@
             </content>
             <outtro>
                 <!--{{#values}}-->
-                <p><m>{{gname}}({{x}})={{result}}</m></p>
+                <p><m>{{gname}}(x)={{result}}</m>when <m>x={{x}}</m></p>
```
AbbyANoble, line 48:
```suggestion
                <p><m>{{gname}}(x)={{result}}</m> when <m>x={{x}}</m></p>
```

Needed a space!

AbbyANoble:
Is there a way to make it all one line, listing all the x's together, instead of two lines like shown below?

Screenshot 2024-12-18 at 1 45 03 PM

AbbyANoble:
"simple" fix.  Gee, thanks Drew. haha. It does leave a trailing comma though.

Screenshot 2024-12-18 at 2 25 24 PM

siwelwerd:
I think this now hits everything you asked for in #495 @AbbyANoble , take a look and let me know.

source/precalculus/exercises/outcomes/FN/FN2/template.xml
```diff
@@ -38,17 +40,20 @@
                 <p><m>{{gname}}({{xvalue}})={{gx}}</m></p>
             </outtro>
         </knowl>
+        <!-- {{/findy}}-->
+        <!-- {{#findx}}-->
         <knowl>
             <content>
                 <p>For which <m>x</m>-values in the interval <m>{{interval}}</m> does <m>{{gname}}(x)={{result}}</m>?
```
AbbyANoble, line 47:
```suggestion
                <p>For which <m>x</m>-value(s) in the interval <m>{{interval}}</m> does <m>{{gname}}(x)={{result}}</m>?
```

AbbyANoble: One little tweak. But looks good to me! Thanks @siwelwerd!

I don't have the review option (perhaps because I did the original PR?), so alerting @StevenClontz to review and approve.

siwelwerd:
@AbbyANoble Please look at this and see if you are happy.  If so, leave a comment, I don't think you can review since you opened the PR.

AbbyANoble:
@siwelwerd Looks good! Please merge.

siwelwerd (approved):

Issue #477, opened 2024-12-17
format command in Octave


jbunn3:
I am suggesting that we find a location in the Linear Algebra textbook (maybe LE2 ?) to introduce the format command as an option in Octave. I find that students get a bit frustrated in later sections with messy matrices that don't align well in Octave. Running `format rat`, for example, will align the columns and convert to fractions. Running `format short` uses 5 significant figures but also aligns the columns. But, if there is a better tool, I'm open to it.

jbunn3:
I got my information from [here](uhttps://docs.octave.org/v4.0.1/Terminal-Output.html#:~:text=The%20format%20command%20offers%20some,through%20the%20normal%20echoing%20mechanism.&text=Display%20the%20value%20of%20x%20.&text=Note%20that%20the%20output%20from,formatted%20output%20in%20a%20string.rl)

siwelwerd:
An option might be to pre-load this as input into some of the sage cells to scaffold this.  Syntax is something like this:

```
<sage language="octave">
<input>rref([2,-3,1;3,0,6])</input>
</sage>
```
And here's how that showed up in-text: [Activity 1.2.19](https://library.tbil.org/2024/linear-algebra/LE2.html#LE2-4-19)

PR #507, opened 2024-12-18
promote `format rat` to LE2

StevenClontz:
inspired by #477 and #479

### DIFF
--- a/source/linear-algebra/source/01-LE/02.ptx
+++ b/source/linear-algebra/source/01-LE/02.ptx
@@ -864,7 +864,9 @@ rref(A)
     </task>
     <task>
         <p>
-Add the line <c>format rat</c> to the top of the cell and re-evaluate
+Add the line <c>format rat</c>
+(short for <q>format using rational numbers</q>)
+to the top of the cell and re-evaluate
 to see nice fractions in place of the decimal approximations.
         </p>
     </task>
--- a/source/linear-algebra/source/01-LE/02.ptx
+++ b/source/linear-algebra/source/01-LE/02.ptx
@@ -824,66 +824,74 @@ Can <m>\RREF(A)</m> be used to find <m>\RREF(B)</m>?
 </activity>
 
 <activity estimated-time='10' xml:id="LE2-rref-tech">
-    <statement>
+    <introduction>
         <p>
 Free browser-based technologies for mathematical computation
 are available online.
         </p>
-<ul>
-<li><p>
-     Go to <url href="https://sagecell.sagemath.org/"/>.
-    </p>
-</li>
-<li>
-    In the dropdown on the right, you can select a number of different languages.
-    Select "Octave" for the Matlab-compatible syntax used by this text.
-</li>
-<li>
-    <p>Type <c>rref([1,3,2;2,5,7])</c> and then press the <kbd>Evaluate</kbd> button to compute the <m>\RREF</m> of 
-      <m>\left[\begin{array}{ccc} 1 &amp; 3 &amp; 2 \\ 2 &amp; 5 &amp; 7 \end{array}\right]</m>.
-    </p>
-</li>
-<li>
-    <p>
-Now try using whitespace to write out the matrix and compute <m>\RREF</m> instead:
-    </p>
-    <program>
-        <input>
-A = [1 3 2
-     2 5 7]
-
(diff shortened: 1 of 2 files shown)


jbunn3:
I think this looks really nice. I didn't initially go that route, since I didn't know if it would be necessary that early, or if it would be a distraction at that stage. However, since there's already [Activity 1.2.19](https://tbil.org/preview/linear-algebra/LE2.html#LE2-rref-tech), I think it fits nice there with the scaffolding Drew suggested.

StevenClontz:
Exactly, this is the activity I'm adapting. Still a work in progress..

source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -824,66 +824,74 @@ Can <m>\RREF(A)</m> be used to find <m>\RREF(B)</m>?
...
-when RREFs need to be computed.
-    </p>
+            </input>
+        </program>
+    </task>
+    <task>
         <p>
-Try this out to compute
-<m>\RREF\left[\begin{array}{cc|c} 2 &amp; -3 &amp; 1 \\ 3 &amp; 0 &amp; 6 \end{array}\right]</m>.
+Add the line <c>format rat</c> to the top of the cell and re-evaluate
```
StevenClontz, line 867:
```suggestion
Add the line <c>format rat</c>
(short for <q>format using rational numbers</q>)
to the top of the cell and re-evaluate
```

source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -824,66 +824,76 @@ Can <m>\RREF(A)</m> be used to find <m>\RREF(B)</m>?
...
+Add the line <c>format rat</c>
+(short for <q>format using rational numbers</q>)
+to the top of the cell and re-evaluate
+to see nice fractions in place of the decimal approximations.
         </p>
-    </statement>
+    </task>
+    <task>
+        <p>
+Visit the HTML version of this activity via <c>TBIL.org</c> to use
```
StevenClontz, line 875:
```suggestion
Visit the HTML version of this activity via <url href="https://TBIL.org/"/> to use
```

source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -824,66 +824,76 @@ Can <m>\RREF(A)</m> be used to find <m>\RREF(B)</m>?
...
+        <p>
+Visit the HTML version of this activity via <url href="https://TBIL.org/"/> to use
+the code cells embedded for your convenience below.
+        </p>
+    </task>
 </activity>
 
 <sage language="octave">
-<input>rref([2,-3,1;3,0,6])</input>
+    <input>rref([1,4,6;2,5,7])</input>
```
jbunn3, line 882: The only issue here is if they evaluate the next example before this one, then this one will evaluate using format rat. You could put `format none` on the first, but that defeats the purpose of showing them what a normal input would be. I'm not really sure what to suggest here. Maybe we just leave it as is? Alternatively, you could add one last sage insert that has `format none` at the top.

jbunn3 (commented):
The rest of the changes look fine, so just take a look at the one comment I made on the format rat part in 02.ptx.

source/linear-algebra/source/01-LE/02.ptx
StevenClontz:
```suggestion
```

I think the slicker solution is to just remove it; I'm not sure it really adds much value if they already did the one liner above.

source/linear-algebra/source/01-LE/02.ptx
```diff
@@ -824,66 +824,75 @@ Can <m>\RREF(A)</m> be used to find <m>\RREF(B)</m>?
...
+(short for <q>format using rational numbers</q>)
+to the top of the cell and re-evaluate
+to see nice fractions in place of the decimal approximations.
         </p>
-    </statement>
+    </task>
+    <task>
+        <p>
+Visit the HTML version of this activity via <url href="https://TBIL.org/"/> to use
+the code cells embedded for your convenience below.
```
StevenClontz, line 876:
```suggestion
the code cell embedded for your convenience below.
```

jbunn3 (approved):
Looks good to me.

PR #511, opened 2024-12-19
Answer added to 1.5 Limits with Infinite Inputs (LT5) except 1.58 (b)

mlamich1:


source/calculus/source/01-LT/05.ptx
```diff
@@ -733,6 +873,11 @@
                         Use the data about the initial temperature to find the value of the parameter <m>a</m> in the model <m> Q(t) = a \, b^t + c</m>.
                     </p>
                 </statement>
+                 <answer>
+                    <p>
+                        <m> Q(t) = 100 (0.9)^t + 72 </m>.
```
cg2wilson, line 878: This should be 28(0.9)^t + 72

source/calculus/source/01-LT/05.ptx
```diff
@@ -633,6 +738,11 @@
                         Then, prove that your guess is right using algebra.
                     </p>
                 </statement>
+                <answer>
+                    <p>
+                        <m> f(x)= 2</m>
```
cg2wilson, line 743: Should this be y = 2?

source/calculus/source/01-LT/05.ptx
```diff
@@ -660,6 +775,11 @@
                         Come up with the formula of a rational function that has horizontal asymptote <m>y=3</m>.
                     </p>
                 </statement>
+                <answer>
+                    <p>
+                        <m> \displaystyle \lim_{x \rightarrow \infty }\frac{3x^2 -x +1 }{x^2+2 }</m>
```
cg2wilson, line 780: I think either we need to say "here is one possible answer" or leave no answer here, since the prompt is to construct a function.

cg2wilson (changes requested):

source/calculus/source/01-LT/05.ptx
StevenClontz:
```suggestion
                        <m> Q(t) = 28(0.9)^t + 72 </m>.
```

source/calculus/source/01-LT/05.ptx
StevenClontz:
```suggestion
                        <m> y= 2</m>
```

source/calculus/source/01-LT/05.ptx
StevenClontz:
```suggestion
                        One possible answer: <m> \displaystyle \lim_{x \rightarrow \infty }\frac{3x^2 -x +1 }{x^2+2 }</m>
```

StevenClontz:
@cg2wilson FYI for next time, when reviewing there's a button you can press to directly suggest changes that can be quickly implemented:

image

PR #519, opened 2024-12-19
Answers added to 1.6 Limits with Infinite Outputs except 1.6.14 (d) and 1.6.15(d)

mlamich1:


cg2wilson (changes requested):
1.6.1: Shouldn't the answer be B. The limit is positive infinity? This keeps the answer consistent with the definition of a vertical asymptote from below.

1.6.15(b): The horizontal asymptote should be y=2

Issue #396, opened 2024-11-03
Typo in AT3 Video 


jkostiuk:
![Screenshot 2024-11-03 at 9 52 40 AM](https://github.com/user-attachments/assets/7868cf0a-fb02-4ac8-ae09-0a6d38238854)
A student found the following typo in a video. 

As a stop-gap, perhaps we can add a note to the video.

jkostiuk:
T(2,1,3,0) ought to be [14,-5,9]

it's not consequential to the outcome because it was just as part of an example, but it gave a few students pause when working along with it.

siwelwerd:
Stop-gap has been implemented, but leaving this open until someone records a new video that we can link.

PR #520, opened 2024-12-19
issue #396

jkostiuk:


siwelwerd:
I didn't check the computation, but the codomain of the example is $\mathbb{R}^3$, so is that zero extra?

jkostiuk:
thanks for the catch @siwelwerd

siwelwerd (approved):

Issue #431, opened 2024-11-18
Log/Exponential properties


AbbyANoble:
We note that $\log_b b^x = x$ and $\log_b 1 =0$ in [Remark 5.3.9](https://library.tbil.org/2024/precalculus/EL3.html#EL3-3-11).

I think we should also point out that $b^{\log_b x} =x$. Maybe this and $\log_b b^x = x$(along with their natural log counterparts) should be together in a remark pointing out that they are a result of logs and exponentials being inverse functions. 

Then $\log_b 1 =0$ and $\ln 1 =0$ could be pointed out too. Not sure what they are officially called? Logs involving 1?

Editing to add that we also need to point out that $\log_b b =1$ and $\ln e = 1$. These (and the ones in the previous paragraph) can both be discovered through an activity using the concept of logs, but at some point we need to write them down.

PR #523, opened 2024-12-19
addresses #431

jkostiuk:
i took a stab at revising the remark Abby mentioned. I also moved the remark dealing with common/natural logarithm earlier because it made more narrative sense to me.


siwelwerd:
@AbbyANoble Check the latest preview, I merged in main so it has incorporated all of the missing <p> fixes.  Any remaining errors you spot we should fix here.

AbbyANoble (approved):

siwelwerd:
So if you look over there now, at the bottom, you see a button for "Resolve Conflicts"

image

I'll do this momentarily, but that button opens up a merge editor where we can reconcile the two branches, and merge the just updated main branch (with Jordan's fix) into that branch over there.

PR #525, opened 2024-12-19
Fixing EQ7

### LINKED ISSUE #562: EQ7 reflections
### Discussed in https://github.com/TeamBasedInquiryLearning/library/discussions/295

<div type='discussions-op-text'>

<sup>Originally posted by **AbbyANoble** September  4, 2024</sup>
Overall, this did not go well today. I felt I had to talk a lot to explain what was going on because there is very little inquiry in the section.

[Activity 1.7.2](https://library.tbil.org/2024/precalculus/EQ7.html#quadratic-inequality-coeff-greater-than) went okay, but depended on looking at a graph. Which is fine, but my students weren't super comfortable with the graphs. (And they learn to graph them later in PR1). We talked about the idea of being above, on, or below the axis. And it did help to reinforce interval notation.

But then [Definition 1.7.3](https://library.tbil.org/2024/precalculus/EQ7.html#def-sign-chart) tells what a sign chart is and then is never mentioned again. I'm also not sure why shading is mentioned as ambiguous in this definition. And why it is "often decorated with a + or -".  And how do we get the + or - ? Would be helpful to have an activity about making a sign chart. Especially to lead them to the factored form being easier because then you can test just for the sign instead of the actual value. 

[Activity 1.7.5](https://library.tbil.org/2024/precalculus/EQ7.html#quadratic-inequality-coeff-less-than-0) asks the students to solve an inequality algebraically for the first time, but they aren't told what that means or how to approach it. Or how to test values and make the sign chart. 

In [Activity 1.7.8(a)](https://library.tbil.org/2024/precalculus/EQ7.html#activity-rational-inequality-1-2) it would help to have them rewrite the inequality where all terms are on one side. My students could see that B or C could both potentially be helpful. But there was confusion over keeping the inequality sign the correct direction. (It "feels" better to move everything to the left. But it's okay to move things to the right as long as you know how to interpret something like  0 > some expression and that it's the same a some expression < 0.)

Is partition value a standard term? Referencing [Definition 1.7.9](https://library.tbil.org/2024/precalculus/EQ7.html#def-partition-values). I feel like I've called them boundary points. Or critical points (though I guess technically those generally deal with the derivative). Either way, maybe this should be introduced earlier to clarify sign charts.

There also needs to be some sort of explanation/motivation for why we can't solve the rational inequalities like we did with rational equations by multiplying through to clear fractions. It's not obvious why that doesn't work. Motivation for getting one rational expression.

Honestly, I just abandoned the activity book except picking out the examples so I didn't have to make them up on the spot. I mostly lectured, with some bouncing around and getting them to finish the problems. 
</div>

tdegeorge:
Closes #562


tdegeorge:
I changed the majority of this section based on @AbbyANoble 's feedback. Please double the graphs as I struggled with some of them! :)

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -159,227 +393,923 @@
...
-          p = arrow((-14,0),(5,0),color="black", width=1, arrowsize=1, aspect_ratio=1,head=2)
-          for i in range(-14,5):
-              p += line([(i,-0.2),(i,0.2)],color="black")
-              p += text(f"${i}$", (i,-0.6),color="black")
-          p += arrow((-12,0),(-15,0),color="#0088ff", width=3)
-          p += line([(-5,0),(2,0)],color="#0088ff", thickness=3)
-          p += text("]", (-12,0), color="#0088ff", fontsize=18)
-          p += text("(", (-5,0), color="#0088ff", fontsize=18)
+          #p += TBILPrecal.inequality_plot(start=-2,end=7,label_endpoints=False)
+          p = arrow((-10,0),(12,0),color="black", width=1, arrowsize=1, aspect_ratio=1,head=2)
```
siwelwerd, line 1091:
```suggestion
          p = arrow((-10,0),(10,0),color="black", width=1, arrowsize=1, aspect_ratio=1,head=2)
```

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -159,227 +393,923 @@
...
+          for i in range(-9,10):
+            p += line([(i,-0.2),(i,0.2)],color="black")
+            p += text(f"${i}$", (i,-0.6),color="black")
+          p += line([(-10,0),(2,0)],color="#0088ff", thickness=3, aspect_ratio=1)
+          p += line([(4,0),(10,0)],color="#0088ff", thickness=3, aspect_ratio=1)
```
siwelwerd, line 1096:
```suggestion
          p += arrow((2,0),(-10,0),color="#0088ff", thickness=3, arrowsize=3)
          p += arrow((4,0),(10,0),color="#0088ff", thickness=3, arrowsize=3)
```
I think it looks better if the shaded part are also arrows

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -159,227 +393,923 @@
...
+        </p>
+      </statement>
+      <answer>
+        <p>
+          <image>
+        <sageplot>
+          <!--<xi:include parse="text" href="../../sage/common.sage"/> -->
+          #p = TBILPrecal.numberline_plot()
+          #p += TBILPrecal.inequality_plot(start=-2,end=7,label_endpoints=False)
+          p = arrow((-10,0),(12,0),color="black", width=1, arrowsize=1, aspect_ratio=1,head=2)
```
siwelwerd, line 1221:
```suggestion
          p = arrow((-10,0),(10,0),color="black", width=1, arrowsize=1, aspect_ratio=1,head=2)
```

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -159,227 +393,923 @@
...
+          for i in range(-9,10):
+            p += line([(i,-0.2),(i,0.2)],color="black")
+            p += text(f"${i}$", (i,-0.6),color="black")
+          p += line([(-10,0),(-2,0)],color="#0088ff", thickness=3, aspect_ratio=1)
```
siwelwerd, line 1225:
```suggestion
          p += arrow((-2,0),(-10,0),color="#0088ff", thickness=3, arrowsize=3)
```

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -13,93 +13,395 @@
...
+  <introduction>
+    <p>
+      Consider the quadratic inequality <me> x^2-6x+5 \gt 0.</me>
+    </p>
+  </introduction>
+
+  <task>
+    <statement>
+      <p>
+        Use a graphing utility to graph the function <m>f(x)=x^2-6x+5</m>. What is the value of <m>f(0)</m>?
```
AbbyANoble, line 39: This is before the functions chapter, so we probably don't need to use function notation.

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -13,93 +13,395 @@
...
+        A <term>quadratic inequality</term> is a mathematical statement that relates a quadratic expression (using <m>\lt, \le, \gt, \ge</m>) to another.
+        </p>
+    </statement>
+  </definition>
+
+  <activity xml:id="graphs-of-quadratic-inequalities">
+
+  <introduction>
+    <p>
+      Consider the quadratic inequality <me> x^2-6x+5 \gt 0.</me>
```
AbbyANoble, line 32: To switch to giving them a graph:
```suggestion
      Consider the graph shown below.
       <figure width="100%" xml:id="EQ7-quadratic">
          <image xml:id="EQ7-quadratic-image">
            <description>Graph of a quadratic</description>
            <sageplot>
              f(x) = (x^2-6*x+5)
                p=plot(f, (x, -1, 5), ymin=-7, ymax=10, color='blue', thickness=3)
            </sageplot>
          </image>
        </figure>
```

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -13,93 +13,395 @@
...
-    <p>In <xref ref="EQ5"/> and <xref ref="EQ6"/> we learned how to solve quadratic and rational equations.  In this section, we use these skills
-      to solve quadratic and rational <em>inequalities</em>.
+    <p>In <xref ref="EQ5"/> and <xref ref="EQ6"/> we learned how to solve quadratic and rational equations.  In this section, we use these skills to solve quadratic and rational <em>inequalities</em>.
     </p>
   </remark>
 
+  <definition xml:id="def-quadratic-inequality">
+    <statement>
+      <p>
+        A <term>quadratic inequality</term> is a mathematical statement that relates a quadratic expression (using <m>\lt, \le, \gt, \ge</m>) to another.
```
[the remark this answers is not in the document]
siwelwerd: Yeah, do we want to say $x^2 \leq x^2$ is a quadratic inequality?  Tonya's definition says yes, but Abby's says no.

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -13,93 +13,395 @@
...
+      <p>
+        C
+      </p>
+    </answer>
+  </task>
+  </activity>
+
+  <remark>
+    <p>
+      From <xref ref="graphs-of-quadratic-inequalities"/>, we saw that a function could have <m>y</m>-values that are positive, negative, or zero, which can then help us find values of <m>x</m> to solve inequalities. Let's now look at how we can solve inequalities using algebra.
```
AbbyANoble, line 173:
```suggestion
      From <xref ref="graphs-of-quadratic-inequalities"/>, we saw that a graph could have <m>y</m>-values that are positive, negative, or zero, which can then help us find values of <m>x</m> to solve inequalities. Let's now look at how we can solve inequalities using algebra.
```

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -13,93 +13,395 @@
...
+            <li><m>(x-1)(x+3) \lt 7</m></li>
+            <li><m>-4x+3 \ge 10</m></li>
+            <li><m>2x^2-7x+3 \gt 0</m></li>
+            <li><m>5x-1 \le 4x</m></li>
+          </ol>
+        </p>
+      </statement>
+      <answer>
+        <p>
+          A and D
```
AbbyANoble, line 198:
```suggestion
          A and C
```

source/precalculus/source/01-EQ/07.ptx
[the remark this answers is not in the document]
AbbyANoble: @siwelwerd - Does the pretext here look right for my list? I did it in the editor rather than firing up a codespace.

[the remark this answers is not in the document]
siwelwerd: Yep, if @tdegeorge accepts this we can double check the preview.

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -13,93 +13,395 @@
...
+    <p>
+      Notice from <xref ref="intro-to-quadratic-inequality"/>, a solution to a quadratic inequality is a real number that will produce a true statement when substituted for the variable. Quadratic inequalities can have infinitely many solutions, one solution, or no solutions. 
+    </p>
+  </remark>
+
+  <activity xml:id="graphing-solutions-of-quadratics-on-number-line">
     <introduction>
-      <p>Consider the quadratic inequality <me>x^2-4x-32\gt0</me>.</p>
+      <p>
+        Let's look at how we can determine the solutions to a quadratic inequality using a number line. Consider the quadratic inequality <me>x^2-4x-32\ge 0</me>.
```
AbbyANoble, line 258:
```suggestion
        Let's look at how we can algebraically determine the solutions to a quadratic inequality using a number line. Consider the quadratic inequality <me>x^2-4x-32\ge 0</me>.
```

source/precalculus/source/01-EQ/07.ptx
```diff
@@ -13,93 +13,395 @@
...
       <statement>
-        <p>Use a graphing utility to graph the function <m>f(x)=x^2-4x-32</m>.  Which part of the graph represents where <m>x^2-4x-32\gt0</m>?</p>
+        <p>
+          What is the factored form of the quadratic (left-hand side of the inequality)?
+          <ol marker="A." cols="2">
+            <li><m>(x-16)(x+2)</m></li>
+            <li><m>(x-4)(x+8)</m></li>
+            <li><m>(x+4)(x-8)</m></li>
+            <li><m>(x-4)(x-8)</m></li>
+          </ol>
```
AbbyANoble, line 270: I think we could probably combine this and the next part into one since they should know how to solve quadratic equations. Maybe something like "We'll start by solving the corresponding quadratic _equation_ x^2-4x-32=0. What are the solution(s)?

source/precalculus/source/01-EQ/07.ptx
AbbyANoble: Coming back to this, maybe we should move this section? If we make it PR7 instead, then they would do it after they have solved the corresponding equations, learned about functions in general, learned more specific stuff about polynomial and rational functions, and graphed them. Or maybe this makes sense as two sections? Polynomial inequalities then Rational inequalities?

source/precalculus/source/01-EQ/07.ptx
tdegeorge: So, if we are moving this section, do we want students to graph it or provide the graph for them?

AbbyANoble: I'd still provide it to cut out some of the time that would be spent graphing it? Then we can all look at the same thing. (I'm finding that I can't get through many of our sections in a 75 minute class period!)

tdegeorge:
@siwelwerd  - I believe I addressed all the comments/suggestions made for this section. Just double check the images in some of the answers (like the number lines). In codechat, the images would not show, so I am not entirely sure they are working/correct. Other than that, this should be ready to split into the two sections we talked about earlier.

siwelwerd:
The graphs look good to me. I went ahead and updated the number lines to use our new library functions, it's way cleaner to read in the code now.

siwelwerd (approved):

PR #526, opened 2024-12-20
Progress on XML validation

StevenClontz:


StevenClontz:
Still lots more validation to be done (and possibly more exceptions to add to our custom schema), but I think this is working correctly now.

StevenClontz:
CC @siwelwerd

PR #560, opened 2025-01-19
Addresses #558 using example environment.

### LINKED ISSUE #558: Tasks not in appropriate parent element
Seeing this warning building the Calculus:

```
warning: * PTX:DEPRECATE: (2017-08-04) the "task" element is no longer used as a child of a top-level division, but is instead being used to divide the other "project-like" elements.  It can be replaced by a functional equivalent: "project", "activity", "exploration", or "investigation" (4 times)
*              located within: "Sample Problem" (title)
*              located within: "Sample Problem" (title)
*              located within: "Sample Problem" (title)
*              located within: "Sample Problem" (title)
* --------------
```

It looks like the sample problem at the end of the new PS5 section needs to get wrapped up in a `<activity>` or something, the `<task>`s can't be children of a subsection

jbunn3:
Closes #558, Example environment added to sample problem.

### DIFF
--- a/source/calculus/source/09-PS/05.ptx
+++ b/source/calculus/source/09-PS/05.ptx
@@ -181,12 +181,12 @@
                 <p>
                     Calculate the derivatives <m>f'(x)</m>, <m>f''(x)</m>, <m>f'''(x)</m>, and <m>f^{(4)}(x)</m>. 
                 </p>
+            </statement>
                 <answer>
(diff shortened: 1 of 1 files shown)


siwelwerd (approved):

StevenClontz:
Thanks @jbunn3 and @siwelwerd!

PR #573, opened 2025-01-30
Add logo

StevenClontz:


StevenClontz:
Ideally we'd use the logo somewhere, but I want to revisit upstream the design of the generic PreTeXt-CLI site to support more customizations, then use those features here.

StevenClontz:
Actually, I should add these to the books...

siwelwerd (approved):

PR #578, opened 2025-02-04
fn3 updates

AbbyANoble:
A few things in here:

1 - Made a lot of the answer choices 4 columns instead of two to save some space and to go with our general use of 4 columns when answers are short. This was helpful because there are large graphs and a lot of scrolling between the graphs and the questions.

2 - In 2.3.20 -- Changed some of the = to approximately. In particular, the global min looks like it occurs at a little less than x=2, not at 2.

### DIFF
--- a/source/precalculus/source/02-FN/03.ptx
+++ b/source/precalculus/source/02-FN/03.ptx
@@ -576,7 +576,7 @@
           <p>
             At approximately what value of <m>x</m> is there a global minimum?
             <ol marker= "A." cols="4">
-              <li> <m>x=\approx -4</m> </li>
+              <li> <m>x\approx -4</m> </li>
               <li> <m>x\approx -2</m> </li>
               <li> <m>x\approx 2</m> </li>
               <li> <m>x\approx 5</m> </li></ol></p>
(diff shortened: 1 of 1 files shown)


source/precalculus/source/02-FN/03.ptx
```diff
@@ -574,12 +574,12 @@
...
           <p>
-            At what value of <m>x</m> is there a global minimum?
-            <ol marker= "A." cols="2">
-              <li> <m>x=-4</m> </li>
-              <li> <m>x=-2</m> </li>
-              <li> <m>x=2</m> </li>
-              <li> <m>x=5</m> </li></ol></p>
+            At approximately what value of <m>x</m> is there a global minimum?
+            <ol marker= "A." cols="4">
+              <li> <m>x=\approx -4</m> </li>
```
tdegeorge, line 579: delete equals sign

tdegeorge (commented):
Looks great - I only saw one place where we need to delete the equals sign. There is probably a fast way for me to do that, but I am not sure how yet.... :)

source/precalculus/source/02-FN/03.ptx
tdegeorge, line 579:
```suggestion
              <li> <m>x\approx -4</m> </li>
```

tdegeorge (approved):

PR #583, opened 2025-02-05
LE3: more accurate condition for infinitely-many solutions

StevenClontz:


source/linear-algebra/source/01-LE/03.ptx
```diff
@@ -306,7 +306,7 @@ and we may write
           <li>
               <p>
 Otherwise, the system given by the <m>\RREF</m> matrix must <em>not include a <m>0=1</m> contradiction</em>
-while <em>including at least one equation with multiple variables</em>. This means it is
+while <em>at least one variable is not required to equal a specific value</em>. This means it is
```
StevenClontz, line 309: Original was mistaken: could have infinitely-many solutions given a missing variable, in which case every equation would still have a single variable.

PR #585, opened 2025-02-05
EQ5 tweaks

### LINKED ISSUE #565: EQ5 Tweaks
### Discussed in https://github.com/TeamBasedInquiryLearning/library/discussions/245

<div type='discussions-op-text'>

<sup>Originally posted by **AbbyANoble** August 21, 2024</sup>
- Reword [Activity 1.5.2](https://library.tbil.org/2024/precalculus/instructor/EQ5.html#determine-quadratic-equation) to indicate that there is more than one answer.

- [Activity 1.5.4(a)](https://library.tbil.org/2024/precalculus/instructor/EQ5.html#applying-zero-product-property-2) - I'd suggest changing part (d) so that it's not equal to 0. That way they aren't just looking for factored pieces, but factored pieces that are set equal to 0. Also, changing answer choice B to match the question they are doing in part b would be good. Some of my students were tripped up in part b because the new quadratic was so close to the one in part a answer choice B.

- My students did not remember how to factor for [Activity 1.5.6](https://library.tbil.org/2024/precalculus/instructor/EQ5.html#solve-unfactored-equation). They mostly did guess and check to narrow down to the right answer for part (a). Which is fine here, I guess. So maybe add factoring when a isn't 1 to the RAT. Or switch this one so that a=1. Then they can practice the idea without factoring holding them back as much. [Activity 1.5.7(c)](https://library.tbil.org/2024/precalculus/instructor/EQ5.html#solve-quadratic-equations-practice-4) also gave them trouble. We talked about how to factor it, but usually I would say that if they don't see how to factor it fairly quickly to just use quadratic formula. But we weren't there yet.

- In [Activity 1.5.7](https://library.tbil.org/2024/precalculus/instructor/EQ5.html#solve-quadratic-equations-practice), it would be nice to have one where a=1 and it's set up in standard form, maybe between b and c. 
</div>

AbbyANoble:
Closes #565

wrote a lot of notes in #565 about all these changes.

### DIFF
--- a/source/precalculus/source/01-EQ/05.ptx
+++ b/source/precalculus/source/01-EQ/05.ptx
@@ -740,7 +740,7 @@
         What is the solution(s) to <m>2x^2-13=7x</m>?
         <ol marker= "A." cols="1">
           <li><m>x=\dfrac{7+\sqrt{73}}{4}</m> and <m>\dfrac{7-\sqrt{73}}{4}</m></li>
-          <li><m>x=\dfrac{7+\3sqrt{17}}{4}</m> and <m>\dfrac{7-\3sqrt{17}}{4}</m></li>
+          <li><m>x=\dfrac{7+3\sqrt{17}}{4}</m> and <m>\dfrac{7-3\sqrt{17}}{4}</m></li>
           <li><m>x=\dfrac{-7+\sqrt{55}}{4}</m> and <m>\dfrac{-7-\sqrt{55}}{4}</m></li>
           <li><m>x=\dfrac{-7+\sqrt{155}}{4}</m> and <m>\dfrac{-7-\sqrt{155}}{4}</m></li></ol>
       </p>
--- a/source/precalculus/source/01-EQ/05.ptx
+++ b/source/precalculus/source/01-EQ/05.ptx
@@ -15,8 +15,11 @@
     <definition xml:id="def-quadratic-equation">
       <statement>
         <p>
-          A <term>quadratic equation</term> is of the form: <me>ax^2+bx+c=0</me> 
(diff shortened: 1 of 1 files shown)


source/precalculus/source/01-EQ/05.ptx
```diff
@@ -726,7 +740,7 @@
         What is the solution(s) to <m>2x^2-13=7x</m>?
         <ol marker= "A." cols="1">
           <li><m>x=\dfrac{7+\sqrt{73}}{4}</m> and <m>\dfrac{7-\sqrt{73}}{4}</m></li>
-          <li><m>x=\dfrac{7+\sqrt{153}}{4}</m> and <m>\dfrac{7-\sqrt{153}}{4}</m></li>
+          <li><m>x=\dfrac{7+\3sqrt{17}}{4}</m> and <m>\dfrac{7-\3sqrt{17}}{4}</m></li>
```
tdegeorge, line 743: Something weird is happening in the preview here. I think the 3 is supposed to be outside the square root stuff - both the positive and negative

tdegeorge (commented):
I think this all looks good (except one part with the square roots - I left a comment). Also left a comment in the discussion about simplifying square roots.

source/precalculus/source/01-EQ/05.ptx
siwelwerd:
```suggestion
          <li><m>x=\dfrac{7+3\sqrt{17}}{4}</m> and <m>\dfrac{7-3\sqrt{17}}{4}</m></li>
```

tdegeorge:
Didn't want to close this yet until we talk about the square roots.

tdegeorge (approved):

PR #602, opened 2025-02-14
Fix reuse of variable

### LINKED ISSUE #597: Task 2.1 in Checkit LF1 Version 3 has incorrect answer
Graph goes through (0,2) and (4,-3).  Answer says slope is -2, rather than -5/4.

siwelwerd:
Closes #597.  We were reusing the variable 'g' for both the graph and the table, so the slope was calculated from g, then it was overwritten for the table, then the graph was produced (with the different g).  Fixed by using a different variable in the table task.


siwelwerd:
While this is labeled as "Checkit", no checkit knowledge is needed to review this. A reviewer just needs to verify that the graph tasks on these now have the correct answers

tdegeorge (approved):

PR #608, opened 2025-02-18
Closes Points vs x-values in Definition 3.1.7 #598

### LINKED ISSUE #598: Points vs x-values in Definition 3.1.7
In the paragraph after the definition, it says "we pick two points, $a$ and $b$, and evaluate the function at those two points.  Following the comment in Issue #580 (Max/Min discussion), would it be more precise/less confusing to refer to $a$ and $b$ as $x$ values?  It seems like that would make it easier to think of the average rate of change as change in $y$ over change in $x$ (slope).

[Definition 3.1.7](https://tbil.org/preview/precalculus/LF1.html#def-average-rate-of-change) 
https://tbil.org/preview/precalculus/LF1.html#def-average-rate-of-change

tdegeorge:
Closes #598

### DIFF
--- a/source/precalculus/source/03-LF/01.ptx
+++ b/source/precalculus/source/03-LF/01.ptx
@@ -172,19 +172,19 @@
     <statement>
       <p>
         The <term>average rate of change</term> of a function on a given interval measures how much the function's value changes per unit on that interval. For a function <m>f(x)</m> on the interval <m>[a,b]</m>, it is calculated by the following expression: 
-        <me>\dfrac{f(b)-f(a)}{b-a}. </me>
-      </p>
-      <p>
-        To calculate the average rate of change for any function <m>f(x)</m>, we pick two <m>x</m>-values, <m>a</m> and <m>b</m>, and evaluate the function at those two <m>x</m>-values. We then find the difference between the <m>y</m>-values and <m>x</m>-values to calculate the average rate of change:
-
         <aside>
           <p>Recall that we can use function notation to describe <m>x</m>- and <m>y</m>-values. <m>f(b)</m>, for instance represents the <m>y</m>-value when plugging in a value for <m>x</m> (or <m>b</m>). </p>
         </aside>
-
-        <me>\dfrac{f(b)-f(a)}{b-a}</me>.
+        <me>\dfrac{f(b)-f(a)}{b-a}. </me>
       </p>
     </statement>
    </definition>
+
+   <remark>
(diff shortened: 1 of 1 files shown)


AbbyANoble:
I agree that saying x-values makes more sense than a point. However, I'm not sure I like our definition here. I think it would make more sense to define the average rate of change on a given interval and point out that it's the end points of the interval that we are using as our two points. As is, it kind of feels like we are saying that to find the average rate of change of a function we just pick two x-values a and b. But average rate of change depends on an interval.

source/precalculus/source/03-LF/01.ptx
```diff
@@ -174,7 +174,7 @@
         An <term>average rate of change</term> of a function calculates the amount of change in one item divided by the corresponding amount of change in another.
       </p>
```
AbbyANoble, line 175:
```suggestion
        The <term>average rate of change</term> of a function on a given interval measures how much the function's value changes per unit on that interval. For a function <m>f(x)</m> on the interval <m>[a,b]</m>, it is calculated by the following expression: 
        <me>\dfrac{f(b)-f(a)}{b-a}. </me>
      </p>
```

source/precalculus/source/03-LF/01.ptx
```diff
@@ -174,7 +174,7 @@
...
       <p>
-        To calculate the average rate of change for any function <m>f(x)</m>, we pick two points, <m>a</m> and <m>b</m>, and evaluate the function at those two points. We then find the difference between the <m>y</m>-values and <m>x</m>-values to calculate the average rate of change:
+        To calculate the average rate of change for any function <m>f(x)</m>, we pick two <m>x</m>-values, <m>a</m> and <m>b</m>, and evaluate the function at those two <m>x</m>-values. We then find the difference between the <m>y</m>-values and <m>x</m>-values to calculate the average rate of change:
```
AbbyANoble, line 177: Then this part could be a remark instead of a definition? Something like:

Remark: Notice that to calculate the aroc over an interval [a,b], we are using the two endpoints of the interval, namely (a,f(a)) and (b,f(b)).

tdegeorge:
Good catch about the interval!

I believe I fixed the changes. I did leave the aside in the definition as I think it is probably a good idea to have a recall with that notation. But I did take out the second part and made that remark. This should be good to go...unless I missed something!

siwelwerd (approved):

PR #609, opened 2025-02-18
m-and-ns in EV2

jkostiuk:
The main fact in EV2 has the number of vectors as n as well as the dimension of the ambient space. I have changed it so that it about n vectors in R^m and I chose this convention so that it's an mxn matrix that we ultimately consider. 

Given this change, I also changed the n to an m in EV2-n-vectors-span (just to not throw folks off). I didn't change the label of this fact though so that we don't have to search for where it's used.

### DIFF
--- a/source/linear-algebra/source/02-EV/02.ptx
+++ b/source/linear-algebra/source/02-EV/02.ptx
@@ -148,7 +148,7 @@ since <m>\IR^1=\setBuilder{cx}{c\in\IR}</m>.
     </statement>
 </activity>
 
-<fact xml:id="EV2-n-vectors-span" >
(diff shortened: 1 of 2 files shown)


StevenClontz (changes requested):
I think it's pretty safe to do a project-wide find/replace for `EV2-n-vectors-span` with `EV2-m-vectors-span`.

Should probably also update the learning outcome at https://github.com/TeamBasedInquiryLearning/library/blob/main/source/linear-algebra/source/02-EV/outcomes/02.ptx

StevenClontz:
image

StevenClontz (approved):

PR #612, opened 2025-02-18
Kathy first draft for TR1 (rebased)

StevenClontz:
Closes #357
Closes #362
Closes #615

PR #613, opened 2025-02-18
update PreTeXt 2.13.5

StevenClontz:


StevenClontz:
@siwelwerd yell at me for self-merging if this breaks (or you don't like the new style) - I need ptx updates to help with #612

PR #621, opened 2025-02-24
establish independence as "solutions are unique"

StevenClontz:


StevenClontz:
Oh and we should probably an activity about the special homogeneous case (since we know a solution exists, it must be unique).

StevenClontz:
[Activity 2.4.5](https://2ec688ab.tbil.pages.dev/preview/linear-algebra/instructor/EV4#EV4-4-5) is the key reworked activity to consider here.

Issue #513, opened 2024-12-19
FN1 updates


AbbyANoble:
### Discussed in https://github.com/TeamBasedInquiryLearning/library/discussions/308

<div type='discussions-op-text'>

<sup>Originally posted by **AbbyANoble** September 11, 2024</sup>
[Activity 2.1.4](https://library.tbil.org/2024/precalculus/FN1.html#FN1-exploring-relations-1)  - 2.1.6 go together and my students did well with them, especially in realizing that it might not be necessary to write things twice. I would recommend one additional exercise to go with this group though. 2.1.4 is one-to-one, 2.1.5 is not a function and also has repeated y values, and 2.1.6 is function but just isn't one-to-one. It might be good to throw in another non-function but no repeated y values. To help point out that it's because of the x-values having two different outputs.

I'd also recommend something to motivate why we care that an input has two outputs but not that the same output happened to come from two different inputs. I like to do this with a restaurant menu. Show them two potential menus and ask which one makes more sense or doesn't have any discrepancies if you were trying to order. Something like:

Menu A 

- Nuggets $2
- Burger $4
- Fries $2
- Nachos $5

Menu B

- Nuggets $2
- Burger $4
- Fries $3
- Nuggets $5

So the first one is a fine menu. We can accept that two different foods have the same price. But the second is a problem. How much do nuggets cost? Who knows?</div>

siwelwerd:
Remaining todo:

> The end of this section has some questions about finding domain and range of a function from the equation. I think one more set of fluency problems would be good. Questions that include some denominators and/or square roots. And I think this should be added as a CheckIt problem.

curtisemartin:
The domain and range part of this discussion was challenging for my students to pick up.  
I also got a bit ahead of myself when giving some similar type problems (find the domain) by giving them a combination of denominators and square roots (before we had made it to FN5 to talk about function combos).

PR #622, opened 2025-02-25
FN1 fluency addition

AbbyANoble:
Addresses some of #513. I added a few domain fluency problems at the end. Once we like the choices, we can add something similar to CheckIt.

### DIFF
--- a/source/precalculus/source/02-FN/01.ptx
+++ b/source/precalculus/source/02-FN/01.ptx
@@ -774,59 +774,60 @@
 <activity xml:id="FN1-determine-domain-from-equation-square-root">
   <introduction>
     <p>
-      It's important to be able to determine the domain of any equation, especially when thinking about functions. Answer the following questions given the equation <m>y=\sqrt{x}</m>.
+      It's important to be able to determine the domain of any equation, especially when thinking about functions. Answer the following questions given the equation <m>y=\sqrt{x-2}</m>.
     </p>
   </introduction>
   <task>
     <statement>
     <p>
-      Which of the following values of <m>x</m> would give an error (if any)?
+      Which of the following values of <m>x</m> would cause <m>y</m> to be undefined (if any)?
       <ol marker= "A." cols="2">
         <li> <m>-2</m> </li>
         <li> <m> 0</m> </li>
+        <li> <m>2</m> </li>
         <li> <m>4</m> </li>
-        <li> <m>-5</m> </li>
         <li> none of the above </li></ol></p>
       </statement>
         <answer>
           <p>
-            A and D
+            A and B
           </p>
         </answer>
       </task>
   <task>
     <statement>
       <p>
-        Based on this information, for which of the following values of <m>x</m> would the equation exist? 
+        Based on this information, for which of the following values of <m>x</m> would <m>y</m> be defined? 
         <ol marker= "A." cols="2">
           <li> <m>-2</m> </li>
           <li> <m> 0</m> </li>
+          <li> <m>2</m> </li>
           <li> <m>4</m> </li>
-          <li> <m>-5</m> </li>
           <li> none of the above </li>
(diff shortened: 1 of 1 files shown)


source/precalculus/source/02-FN/01.ptx
```diff
@@ -956,6 +956,49 @@
     </task>
 </activity>
 
+<activity xml:id="FN1-domain">
+    <introduction>
+      <p>
+       Find the domain of each of the following functions. Write your answer in interval notation
```
tdegeorge, line 962:
```suggestion
       Find the domain of each of the following functions. Write your answer in interval notation.
```

tdegeorge:
I saw in the discussion about adding in range. Did we want to add in finding the range of a function from an equation? I don't normally have students find range at this point, but it could be something we have in there and people can choose to use it (or not).

tdegeorge:
Should we change [[Activity 2.1.21](https://c9a98b28.tbil.pages.dev/preview/precalculus/instructor/FN1#FN1-determine-domain-from-equation-square-root) 
https://c9a98b28.tbil.pages.dev/preview/precalculus/instructor/FN1#FN1-determine-domain-from-equation-square-root](url) so that it is not just the square root of x? Maybe change that to the square root of 4x-1 or something different? It might help students make connections to the fluency problem part b in the last activity. Also, there are no examples of the square root in the denominator until the fluency question at the end.

AbbyANoble:
> I saw in the discussion about adding in range. Did we want to add in finding the range of a function from an equation? I don't normally have students find range at this point, but it could be something we have in there and people can choose to use it (or not).

I don't generally do range from an equation either. Maybe we can make that a non-urgent issue if someone wants to tackle it?

AbbyANoble:
> Should we change ([Activity 2.1.21]
> https://c9a98b28.tbil.pages.dev/preview/precalculus/instructor/FN1#FN1-determine-domain-from-equation-square-root) so that it is not just the square root of x? Maybe change that to the square root of 4x-1 or something different? It might help students make connections to the fluency problem part b in the last activity. Also, there are no examples of the square root in the denominator until the fluency question at the end.

Yeah. I'd suggest something like $sqrt{x-2}$ though at first so it's easier to see. I think the first one having a fractional endpoint may be too much to start with! At least for my students.

AbbyANoble:
I changed up the $\sqrt{x}$ problem if you want to check it out.

tdegeorge:
> > Should we change ([Activity 2.1.21]
> > https://c9a98b28.tbil.pages.dev/preview/precalculus/instructor/FN1#FN1-determine-domain-from-equation-square-root) so that it is not just the square root of x? Maybe change that to the square root of 4x-1 or something different? It might help students make connections to the fluency problem part b in the last activity. Also, there are no examples of the square root in the denominator until the fluency question at the end.
> 
> Yeah. I'd suggest something like x − 2 though at first so it's easier to see. I think the first one having a fractional endpoint may be too much to start with! At least for my students.
> 
> > Also, there are no examples of the square root in the denominator until the fluency question at the end.
> 
> Do you think that's something they could put together after seeing that denominators and square roots affect the domain? What if we added in a hint to it rather than a full exercise? I feel like it's getting so long already!

I like the hint idea!

tdegeorge (approved):

tdegeorge:
I think we should probably change the wording so that it matches in the following activities as well ("Which of the following values of would cause y to be undefined (if any)?" rather than saying "error").

And then adding the hint in the last problem would be good I think!

AbbyANoble:
> I think we should probably change the wording so that it matches in the following activities as well ("Which of the following values of would cause y to be undefined (if any)?" rather than saying "error").
> 
> And then adding the hint in the last problem would be good I think!

I changed the wording and added the hint in! Ready to approve and merge if you are good with this!

PR #629, opened 2025-02-25
Common sagemath library across books+banks

StevenClontz:


source/common/sagemath/library.sage
```diff
@@ -30,3 +34,332 @@ class TBIL:
...
+            if label_endpoints:
+                P += text(f"${round(end,ndigits=2)}$", (end,0.6), color="black")
+            if strict_end:
+                P += text(")", (end,0), color="#0088ff", fontsize=18)
+            else:
+                P += text("]", (end,0), color="#0088ff", fontsize=18)
+        return P
+
+    @staticmethod
+    def intervals_from_inequality(inequality, partition_points,undefined_points=[],checkpoints=None):
```
StevenClontz, line 87:
```suggestion
    def intervals_from_inequality(inequality, partition_points,undefined_points=None,checkpoints=None):
        if undefined_points is None: undefined_points = []
```

StevenClontz:
> Running a full build at https://github.com/TeamBasedInquiryLearning/library/actions/runs/13532385581

hmm apparently that doesn't do a full build after all

source/common/sagemath/library.sage
siwelwerd: How is this different?

source/common/sagemath/library.sage
```diff
@@ -30,3 +34,333 @@ class TBIL:
...
+        return P
+
+    @staticmethod
+    def small_rationals(numerators=range(-8,9), 
+                        denominators=[2,3,5],
+                        dictionary=True,
+                        length=1):
+        '''Generates a list or dictionary of unique rational numbers with small numerators and denominators. 
+        For a dictionary, keys are the rationals and values are the denominators.'''
+        fulldict = {m/n:n for m in numerators for n in denominators if m%n !=0}
```
StevenClontz, line 183:
```suggestion
                        denominators=None,
                        dictionary=True,
                        length=1):
        '''Generates a list or dictionary of unique rational numbers with small numerators and denominators. 
        For a dictionary, keys are the rationals and values are the denominators.'''
        if denominators is None: denominators = [2,3,5]
        fulldict = {m/n:n for m in numerators for n in denominators if m%n !=0}
```

source/common/sagemath/library.sage
```diff
@@ -30,3 +34,333 @@ class TBIL:
...
+    @staticmethod
+    def small_irrationals(rational_part=range(-8,9), 
+                          irrational_part=[2,3,5,6,7,8],
+                          denominators=[i for i in range(-5,6) if i != 0],
+                          dictionary=True,
+                          length=1,
+                          full_list=False):
+        '''Generates a list or dictionary of uniqe irrational numbers of the form (a+sqrt(b))/c. 
+        For a dictionary, keys are tuples of rationals and their conjugate, and values are the denominators.'''
+        fulldict = { ((a+sqrt(b))/c,(a-sqrt(b))/c):c for a in rational_part for b in irrational_part for c in denominators}
```
StevenClontz, line 198:
```suggestion
                          irrational_part=None,
                          denominators=None,
                          dictionary=True,
                          length=1,
                          full_list=False):
        '''Generates a list or dictionary of uniqe irrational numbers of the form (a+sqrt(b))/c. 
        For a dictionary, keys are tuples of rationals and their conjugate, and values are the denominators.'''
        if irrational_part is None: irrational_part = [2,3,5,6,7,8]
        if denominators is None: denominators=[i for i in range(-5,6) if i != 0]
        fulldict = { ((a+sqrt(b))/c,(a-sqrt(b))/c):c for a in rational_part for b in irrational_part for c in denominators}
```

source/common/sagemath/library.sage
StevenClontz: https://www.pullrequest.com/blog/python-pitfalls-the-perils-of-using-lists-and-dicts-as-default-arguments/

siwelwerd: Tell me again about how Python is great because the code is so readable? 🫠

(Thanks for the explainer.)

siwelwerd (approved):

siwelwerd:
I am reasonably confident this will not break things too terribly, so feel free to merge.

source/common/sagemath/library.sage
StevenClontz: Yeah, I'm not sure what advantage there is to assign default arguments at function instantiation rather than function call, but there it is.

StevenClontz:
let's keep an eye on https://github.com/TeamBasedInquiryLearning/library/actions/runs/13533095412 to make sure it deploys

source/common/sagemath/library.sage
siwelwerd: I'm glad you noticed this, because I feel like we have chased a bug like this around before and it took a while to find.

siwelwerd:
> let's keep an eye on [TeamBasedInquiryLearning/library/actions/runs/13533095412](https://github.com/TeamBasedInquiryLearning/library/actions/runs/13533095412) to make sure it deploys

It did not deploy.  I think when including the library, everything needs to be flush left because Python is stupid and cares about whitespace.

Will submit a separate PR fixing that.

PR #630, opened 2025-02-26
answer fix in LF6

AbbyANoble:
The answer is showing up in the student version for Activity 3.6.4 
https://tbil.org/preview/precalculus/LF6.html#LF4-solution-to-system

I think I've fixed it, but the codechat shows the instructor version so I'm not sure!

### DIFF
--- a/source/precalculus/source/03-LF/06.ptx
+++ b/source/precalculus/source/03-LF/06.ptx
@@ -94,7 +94,7 @@
 </definition>
 
 <activity xml:id="LF4-solution-to-system">
-  <introduction> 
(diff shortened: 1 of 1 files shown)


siwelwerd:
> I think I've fixed it, but the codechat shows the instructor version so I'm not sure!

You did!

For future reference, if you need to look at the student version, go to the `codechat_config.yaml` file, and remove the `-instructor` on lines 1 and 2 (so that it just says `precalculus-web` instead of `precalculus-web-instructor`).

siwelwerd (approved):

PR #647, opened 2025-03-04
Add features to TBIL.plot_angle()

siwelwerd:
Based on what @AbbyANoble was trying to do today, I added a few (optional) features to `TBIL.plot_angle()`.  New features:

- A `label_unit_point` argument that controls the appearance of the label at the endpoint. If false, no labels appear.  If a tuple is passed, this will be the x and y components of the label.  Otherwise, the x and y value of the end point are used.
- A `show_triangle` argument that will draw the right triangle
- A `triangle_labels` argument that will label the legs of this triangle.  Passing a tuple allows the user to specify these labels, otherwise the lengths are used for any other true value.

@AbbyANoble were there any other features that might be helpful to add?

Once this is merged I'll update the wiki to document these features.

### DIFF
--- a/source/common/sagemath/library.sage
+++ b/source/common/sagemath/library.sage
@@ -9,7 +9,10 @@ class TBIL:
         show_axes=True,
         show_unit_circle=False,
         show_unit_point=False,
-        show_angle_value=False
+        label_unit_point=False,
+        show_angle_value=False,
+        show_triangle=False,
+        triangle_labels=False,
     ):
         reference_coordinate = (cos(reference_angle),sin(reference_angle))
         end_angle = reference_angle+angle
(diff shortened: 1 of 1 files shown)


AbbyANoble:
Could we add:

- option for the right angle symbol when the triangle is drawn in
- option to label the angle with its measure near the arc, with a degree and radian option
- when drawing a reference angle, include the arc of the original angle, with the option to label it. 
IMG_7721

siwelwerd:
> with a degree and radian option

I missed this part, let me see...

siwelwerd:
`TBIL.plot_angle(pi/3,show_angle_value=True,degree_labels=True)`
image

siwelwerd:
In #653 I added a little helper function to typeset angles as $\frac{2\pi}{3}$ instead of $2\frac{\pi}{3}$, that should be integrated into this, but I don't want to create a merge nightmare.

siwelwerd:
What is the desired behavior if `reference_angle` is nonzero and `show_right_triangle` is `True`?

siwelwerd:
`reference_angle` is maybe a misleading name for the argument

AbbyANoble (approved):

Issue #533, opened 2024-12-20
Update labels on trig graphs


siwelwerd:
In [Section PF1](https://tbil.org/preview/precalculus/instructor/PF1.html) and maybe elsewhere, there are a bunch of trig graphs  with tick marks labeled things like $\frac{1}{2} \pi$.  But in looking at one of the images in Calculus CO-4 I figured out how to get these to typeset as $\frac{\pi}{2}$.  We should copy that idea over to make the other graphs look nicer.

siwelwerd:
Hmm, this is more involved than I thought it would be.

siwelwerd:
Key insight from the [Sage documentation](https://doc.sagemath.org/html/en/reference/plotting/sage/plot/graphics.html): "Custom tick labels can be provided by providing the keyword tick_formatter with the list of labels, and simultaneously providing the keyword ticks with the positions of the labels."

Wired this up into a quick proof of concept

Old image:

![image](https://github.com/user-attachments/assets/52bdbec8-9c0b-402c-b8ab-c754f35a794b)

New image:
![image](https://github.com/user-attachments/assets/a6bf4346-2132-458c-9e48-85dfbf47eda2)

Code:
```
f(x) = sin(x)
p=Graphics()
xmin=-2*pi
xmax=4*pi
delta=pi/4
custom_ticks=[]
custom_tick_labels=[]
for x in [xmin+delta*i for i in [0.. int((xmax-xmin)/delta)]]:
    if denominator(x) != 1:
        custom_tick_labels.append(f"$\\dfrac{{{latex(numerator(x))}}}{{{denominator(x)}}}$")
        custom_ticks.append(x)
    else:
        custom_ticks.append(x)
        custom_tick_labels.append(f"${latex(x)}$")
        
p+=plot(f, (x, xmin, xmax), color='blue', thickness=3, detect_poles=True,ticks=[custom_ticks,0.5], tick_formatter=[custom_tick_labels,"latex"])
show(p)
```

Probably the "right" thing to do is factor this into a function in our common.sage

siwelwerd:
> Probably the "right" thing to do is factor this into a function in our common.sage

Ah, but that's probably blocked by #361

siwelwerd:
This was closed by #670, which subsumed #667

PR #653, opened 2025-03-10
Add trig_plot function

siwelwerd:
See #533 , this library function allows users to easily set labels along the $x$-axis to integer multiples of a rational fraction of pi, as we commonly want to do with graphs of trig functions, but with labels typeset as e.g. $\frac{3\pi}{2}$ instead of $\frac{3}{2}\pi$.  All arguments other than `ticks` are passed along to the underlying plot function, allowing for further customization.

`TBIL.trig_plot(sin(x),(x,0,4*pi),ticks=pi/2,color='red')`
image

### DIFF
--- a/source/common/sagemath/library.sage
+++ b/source/common/sagemath/library.sage
@@ -187,11 +187,10 @@ class TBIL:
         custom_ticks=[]
         custom_tick_labels=[]
         for x in [xmin+delta*i for i in [0.. int((xmax-xmin)/delta)]]:
+            custom_ticks.append(x)
             if denominator(x) != 1:
                 custom_tick_labels.append(f"$\\dfrac{{{latex(numerator(x))}}}{{{denominator(x)}}}$")
-                custom_ticks.append(x)
             else:
-                custom_ticks.append(x)
                 custom_tick_labels.append(f"${latex(x)}$")
                 
         #Default formatting
--- a/source/common/sagemath/library.sage
+++ b/source/common/sagemath/library.sage
@@ -177,6 +177,37 @@ class TBIL:
         return P
 
     @staticmethod
+    def trig_plot(f, *args, **kwds):
+        xmin=-2*pi
+        xmax=4*pi
+        if 'ticks' not in kwds.keys() or type(kwds['ticks']) is not sage.symbolic.expression.Expression:
+            delta=pi/4
+        else:
+            delta=kwds['ticks']
+        custom_ticks=[]
+        custom_tick_labels=[]
+        for x in [xmin+delta*i for i in [0.. int((xmax-xmin)/delta)]]:
+            if denominator(x) != 1:
+                custom_tick_labels.append(f"$\\dfrac{{{latex(numerator(x))}}}{{{denominator(x)}}}$")
+                custom_ticks.append(x)
+            else:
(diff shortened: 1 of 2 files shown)


source/common/sagemath/library.sage
```diff
@@ -176,6 +176,37 @@ class TBIL:
...
+            delta=kwds['ticks']
+        custom_ticks=[]
+        custom_tick_labels=[]
+        for x in [xmin+delta*i for i in [0.. int((xmax-xmin)/delta)]]:
+            if denominator(x) != 1:
+                custom_tick_labels.append(f"$\\dfrac{{{latex(numerator(x))}}}{{{denominator(x)}}}$")
+                custom_ticks.append(x)
+            else:
+                custom_ticks.append(x)
+                custom_tick_labels.append(f"${latex(x)}$")
```
siwelwerd, line 195:
```suggestion
            custom_ticks.append(x)
            if denominator(x) != 1:
                custom_tick_labels.append(f"$\\dfrac{{{latex(numerator(x))}}}{{{denominator(x)}}}$")
            else:
                custom_tick_labels.append(f"${latex(x)}$")
```

source/common/sagemath/library.sage
```diff
@@ -176,6 +176,36 @@ class TBIL:
             end=i["right"], strict_end=i["right_strict"], label_endpoints=False,scale=scale)
         return P
 
+    @staticmethod
+    def trig_plot(f, *args, **kwds):
+        xmin=-2*pi
+        xmax=4*pi
```
siwelwerd, line 182:
```suggestion
        if len(args)==0:
            raise ValueError("Second positional argument should be a tuple of the form (var, min, max) or (xmin, xmax)")
        if type(args[0]) is tuple and len(args[0])==2:
            xmin=args[0][0]
            xmax=args[0][1]
        elif type(args[0]) is tuple and len(args[0])==3:
            xmin=args[0][1]
            xmax=args[0][2]
        else:         
            xmin=0
            xmax=2*pi
```

source/common/sagemath/library.sage
```diff
@@ -176,6 +176,56 @@ class TBIL:
...
+        else:
+            delta=kwds['ticks']
+        custom_ticks=[]
+        custom_tick_labels=[]
+        for x in [xmin+delta*i for i in [0.. int((xmax-xmin)/delta)]]:
+            custom_ticks.append(x)
+            if denominator(x) != 1:
+                custom_tick_labels.append(TBIL.typeset_angle(x))
+            else:
+                custom_tick_labels.append(f"${latex(x)}$")
```
siwelwerd, line 214:
```suggestion
            custom_tick_labels.append(TBIL.typeset_angle(x))
```

StevenClontz (commented):
Looks reasonable. Can we add an example using it to a section somewhere as part of this PR to make sure it works?

siwelwerd:
Yeah, I'll add one to PF1 maybe

siwelwerd:
In lieu of the preview, here's a screenshot of the produced image from 7.1.5(d)

image

StevenClontz (approved):
Thanks for supplying the screenshot. Hoping to sync up with Oscar tomorrow to get this fixed.

PR #657, opened 2025-03-12
AT3 Incorporate Vector Equations

jkostiuk:
I reworked 3.3.5 and 3.3.10 to reinforce the translation of an equation involving a transformation to a vector equation

### DIFF
--- a/source/linear-algebra/source/03-AT/03.ptx
+++ b/source/linear-algebra/source/03-AT/03.ptx
@@ -82,12 +82,12 @@ the set of all vectors that transform into <m>\vec 0</m>?
 <definition xml:id="AT3-definition-kernel">
     <statement>
         <p>
-Let <m>T: V \rightarrow W</m> be a linear transformation, and let <m>\vec{z}</m> be the additive
-identity (the <q>zero vector</q>) of <m>W</m>.  The <term>kernel</term><idx>kernel</idx> of <m>T</m>
+Let <m>T: V \rightarrow W</m> be a linear transformation, and let <m>\vec{0}</m> be the zero vector of <m>W</m>.  
+The <term>kernel</term><idx>kernel</idx> of <m>T</m>
 (also known as the <term>null space</term><idx>null space</idx> of <m>T</m>)
 is an important subspace of <m>V</m> defined by
 <me>
-\ker T = \left\{ \vec{v} \in V\ \big|\ T(\vec{v})=\vec{z}\right\}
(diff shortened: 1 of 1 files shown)


jkostiuk:
copy-pasting comments from an instructor so I don't lose them (haven't yet reflected on it):

_I feel like there could still be stated a Fact: the image is equal to the span of these column vectors. I scaffolded a warmup activity for my students today to try do this. It was something like:
Find a linear combination of T(e1),... that is equal to the vector T(1,3,-0.5,7) that's in the image. And explain why you know these things are equal, without having to actually do all of the computations. This shows everything in this span is a member of the image.
Find a vector in the domain that transforms to the linear combination 4*T(e1)+.... This explains why everything in the image is a member of this span.
I think my students internalized this too well and we had some disagreement about whether there was actually anything to prove here. I tried to highlight that these definitions, of image and span, are a priori totally unrelated, and there has to be an explanation of why they're connected.

There could maybe also be a Fact: the kernel is the solution set of this homogeneous vector equation. Just to spell out the connection between the linear transformation objects and the vector equation/linear system objects_

StevenClontz:
I had the exact thoughts as your TAs when I did this today. Going to try to push a hotfix here to reflect my on the fly changes...

StevenClontz:
image

Note that in this specific example, the answer "infinitely-many solutions" is valid, because the set is dependent (there cannot be exactly one solution). I'm going to tweak the example to avoid this.

StevenClontz:
Apolgies for writing over your work rather than setting up a separate PR in my haste to make some changes between my classes. The latest commit reflects how I ended up pitching things in the second class (or the changes I wish I had made). What do you think?

jkostiuk:
@StevenClontz if you request a review from me, I can make some minor edits to the grammar for your subsequence review.

StevenClontz:
I can't request your review for your own PR. But you can still suggest changes using the "Files changed" tab.

jkostiuk:
y'know @StevenClontz, after being away from this for a long time, I think these activities could still use some work. 

Specifics of any activities in AT3 aside, a core idea I would want my students to get comfortable with is the use of linear transformations in forming/solving equations. That is, I want students to consider questions/statements about equations of the form T(x)=w and be comfortable with translating these into question/statements about vector equations. 

In this PR, activities 3.3.5/remark 3.3.6 i think hit the mark (for me), but 3.3.11 doesn't quite handle this. I'm going to draft a different PR so we don't lose this one and then I'm hoping you and I can sync up live this week to resolve?

siwelwerd:
Is this still needed now that #901 has been merged?

PR #663, opened 2025-03-17
TE3 Draft done

tdegeorge:


tdegeorge:
This was supposed to say TE3 Draft done!

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+    <introduction>
+    <p>
+    Consider the equation <m>(2\sin\theta+\sqrt{3})(\cos\theta-1)=0</m>.
+    </p>
+    </introduction>
+
+    <task>
+    <statement>
+    <p>
+    Refer back to the Zero Product Property. What equations do we solve?
```
siwelwerd, line 129:
```suggestion
    Refer back to the Zero Product Property <xref ref="def-zero-product-property"/>. What equations do we solve?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+            <li> <m>\frac{2\pi}{3}</m> </li>
+            <li> <m>\frac{7\pi}{6}</m> </li>
+            <li> <m>\frac{11\pi}{6}</m> </li></ol>
+          </p>
+        </statement>
+        <answer>
+          <p>
+            A, C, E, and F
+          </p>
+        </answer>
```
siwelwerd, line 557: I don't see the correct answers among the answer choices, and thus the supplied answer is incorrect.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+    <introduction>
+      <p>
+        Consider the equation <m>\sin(2\theta)=\frac{1}{2}</m>. 
+      </p>
+    </introduction>
+
+    <task>
+      <statement>
+        <p>
+          Although we are given <m>\sin(2\theta)=\frac{1}{2}</m>, let's first think about values of <m>\theta</m> that would satisfy <m>\sin\theta=\frac{1}{2}</m>. What values of <m>\theta</m> would satisfy the equation <m>\sin\theta=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```
siwelwerd, line 888: I think it might be better to use a different variable.  Say e.g. $\alpha = 2 \theta$, solve it for $\alpha$, then bring back $\theta$ at the end.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+    </introduction>
+
+    <task>
+    <statement>
+    <p>
+    What first step would you take to isolate <m>\sin\theta</m>?
+    <ol marker= "A." cols="1">
+            <li> Divide both sides by <m>2</m>. </li>
+            <li> Add <m>1</m> to both sides. </li>
+            <li> Subtract <m>2\sin\theta</m> to both sides. </li></ol>
```
AbbyANoble, line 35:
```suggestion
            <li> Subtract <m>2\sin\theta</m> from both sides. </li></ol>
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+        <p>
+          B
+        </p>
+      </answer>
+    </task>
+
+    <task>
+    <statement>
+    <p>
+    If you continue to isolate <m>\sin\theta</m>, what equation would you have left?
```
AbbyANoble, line 53:
```suggestion
    Continue using algebra to isolate <m>\sin\theta</m>. What equation would you have left?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+      <p>
+        Recall that one complete revolution is <m>2\pi</m>, so the positive <m>x</m>-axis can correspond to either an angle of <m>0</m> or <m>2\pi</m> radians (or <m>4\pi</m>, or <m>6\pi</m>, or <m>-2\pi</m>, or <m>-4\pi</m>, etc. depending on the direction of rotation). How can we represent all the possible solutions to the equation if there are <m>n</m> complete revolutions (where <m>n=0, \pm1, \pm2,...</m>)?
+      </p>
+    </statement>
+    <answer>
+      <p>
+        <m>\theta=\frac{\pi}{6}+2\pi(n)</m> and <m>\theta=\frac{5\pi}{6}+2\pi(n)</m>
+      </p>
+    </answer>
+    </task>
```
AbbyANoble, line 104: I think this might need to be its own observation/activity, learning how to do this notation for the general solution. Something like:

"Notice in Activity 8.3.2 we only asked for values of theta on the interval [0,2pi). There are more angles that solve the equation than just the two we found. In fact, any angle coterminal with the either of the angles we found will also be solutions. That means, we can add or subtract any multiple of 2pi to the solutions to find another solution. The notation we use for this is 

$\theta = \dfrac{\pi}{6}+2\pi n$ where $n$ is an integer

and 

 $\theta = \dfrac{5\pi}{6}+2\pi n$ where $n$ is an integer

AbbyANoble: Mainly, I don't think this notation is necessarily something they can come up with on their own, but they need to see it.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+    <remark>
+    <p>
+    One way to solve a trigonometric equation is to isolate the trigonometric term on one side. Once it has been isolated, you can then determine values of <m>\theta</m> that satisfies the equation.
+    </p>
+    </remark>
+
+    <definition xml:id="general-solution">
+      <p>
+        In many cases, you will be asked to find the solution to a trigonometric equation within a given interval (such as <m>[0, 2\pi)</m>). If, however, you are asked to find the <term>general solution</term> of a trigonometric equation (like in <xref ref="intro-to-trig-equations"/>), it will require you to think more broadly (beyond <m>2\pi</m>), where <m>n=0, \pm1, \pm2,...</m> and represents the number of complete revolutions.
+      </p>
```
AbbyANoble, line 116:
```suggestion
      <p>
        In many cases, you will be asked to find the solution to a trigonometric equation within a given interval (such as <m>[0, 2\pi)</m>). If, however, you are asked to find the <term>general solution</term> of a trigonometric equation (like in <xref ref="intro-to-trig-equations"/>), you need to include all solutions. This requires the notation below, which includes all angles coterminal with the solutions in the interval <m>[0, 2\pi)</m>.  If <m> \alpha</m> is a solution to a trigonometric equation, then the general solution would be 
       <me>\theta = \alpha + 2\pi n, n \in \mathbb{Z} </me> 
      where <m>\mathbb{Z} </m> represents the set of integers.
</p>
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+    <introduction>
+    <p>
+    Consider the equation <m>(2\sin\theta+\sqrt{3})(\cos\theta-1)=0</m>.
+    </p>
+    </introduction>
+
+    <task>
+    <statement>
+    <p>
+    Refer back to the Zero Product Property <xref ref="def-zero-product-property"/>. What equations do we solve?
```
AbbyANoble, line 129:
```suggestion
    Refer back to the Zero Product Property <xref ref="def-zero-product-property"/>. What equations do we need to solve?
```

source/precalculus/source/08-TE/03.ptx
AbbyANoble: So now I see the definition below. Maybe what I've just written could be an activity after the definition? And remove part (d) from 8.3.2.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+      <answer>
+        <p>
+          <m>\theta=0+2\pi(n)</m>
+        </p>
+        <p>
+          <m>\theta=\frac{4\pi}{3}+2\pi(n)</m>
+        </p>
+        <p>
+          <m>\theta=\frac{5\pi}{3}+2\pi(n)</m>
+        </p>
```
AbbyANoble, line 206: Do you not require them to write something about n in these? Like our precalc requires students to write

$\theta=\dfrac{4\pi}{3}+2\pi(n), n \in \mathbb{Z}$

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+        <p>
+          <m>\sin\theta-1=0</m> and <m>\csc\theta+2=0</m>
+        </p>
+      </answer>
+    </task>
+
+    <task>
+      <statement>
+        <p>
+          Solve <m>\sin\theta-1=0</m> on the interval <m>[0, 2\pi)</m>. What values of <m>\theta</m> satisfies that equation?
```
AbbyANoble, line 296:
```suggestion
          Solve <m>\sin\theta-1=0</m> on the interval <m>[0, 2\pi)</m>. What values of <m>\theta</m> satisfy that equation?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+      <p>
+        D
+      </p>
+    </answer>
+    </task>
+
+    <task>
+      <statement>
+        <p>
+          Using what you know from part (d), what values of <m>\theta</m> satisfies that equation <m>\csc\theta=-2</m> on the interval <m>[0, 2\pi)</m>?
```
AbbyANoble, line 345:
```suggestion
          Using what you know from part (d), what values of <m>\theta</m> satisfy that equation <m>\csc\theta=-2</m> on the interval <m>[0, 2\pi)</m>?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+      <m>\theta=0, \frac{2\pi}{3}, \frac{4\pi}{3}</m>
+      </p>
+      </answer>
+      </task>
+    </activity>
+
+    <remark>
+      <p>
+        In some cases, it is not easy to isolate the trigonometric term without having to use some other strategies we have learned. The next few activities will help us learn how to apply methods we have used before to solve trigonometric equations.
+      </p>
```
AbbyANoble, line 481:
```suggestion
      <p>
        In some cases, it is not easy to isolate the trigonometric term like we have been doing so far. The next few activities will help us learn how to apply methods we have used before when solving algebraic equations to solve trigonometric equations.
      </p>
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+          <p>
+            <m>\sin^2\theta=\frac{1}{2}</m>
+          </p>
+        </answer>
+      </task>
+
+      <task>
+        <statement>
+          <p>
+            To undo a squared term, you will need to take the square root (recall square root property). Take the square root of <m>\sin^2\theta</m> to get <m>\sin\theta</m>. What does <m>\sin\theta</m> equal?
```
AbbyANoble, line 507: Might help to link to [Definition 1.5.9](https://tbil.org/preview/precalculus/instructor/EQ5.html#def-square-root-property) here.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+          <p>
+            D
+          </p>
+        </answer>
+      </task>
+
+      <task>
+        <statement>
+          <p>
+            Notice there are two solutions <m>\sin\theta</m>, which corresponds to <m>4</m> values of <m>\theta</m> that satisfies the equation <m>\sin\theta=\pm\frac{\sqrt{2}}{2}</m>. What values of <m>\theta</m> satisfies that equation on the interval <m>[0, 2\pi)</m>?
```
AbbyANoble, line 543:
```suggestion
            Notice there are two values of <m>\sin\theta</m> to consider, which corresponds to four values of <m>\theta</m> that satisfy the equation <m>\sin\theta=\pm\frac{\sqrt{2}}{2}</m> on the interval <m>[0, 2\pi)</m>. What are these values of <m>\theta</m>?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+        </answer>
+      </task>
+
+      <task>
+        <statement>
+          <p>
+            Factor out the GCF. What equation do you now have?
+            <ol marker= "A." cols="1">
+            <li> <m>2\sin\theta-\sqrt{2}=0</m> </li>
+            <li> <m>\cos\theta(2\sin\theta=\sqrt{2}\cos\theta)=0</m> </li>
```
AbbyANoble, line 670:
```suggestion
            <li> <m>\cos\theta(2\sin\theta-\sqrt{2}\cos\theta)=0</m> </li>
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+          <p>
+            A and C
+          </p>
+        </answer>
+      </task>
+
+      <task>
+        <statement>
+          <p>
+            Solve each of the equations you got in part (c). What values of <m>\theta</m> satisfies both equations on the interval <m>[0, 2\pi)</m>?
```
AbbyANoble, line 703:
```suggestion
            Solve each of the equations you got in part (c). What value(s) of <m>\theta</m> satisfy both equations on the interval <m>[0, 2\pi)</m>?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+            <li> <m>\frac{\pi}{4}</m> </li>
+            <li> <m>\frac{7\pi}{4}</m> </li>
+            <li> <m>2\pi</m> </li></ol>
+          </p>
+        </statement>
+        <answer>
+          <p>
+            A, C, and D
+          </p>
+        </answer>
```
AbbyANoble, line 717: Solutions are $\dfrac{\pi}{2}$, $\dfrac{3\pi}{2}$, $\dfrac{\pi}{4}$, $\dfrac{3\pi}{4}$.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+    <introduction>
+      <p>
+        Consider the equation <m>2\cos^2\theta-5\cos\theta+3=0</m>.
+      </p>
+    </introduction>
+
+    <task>
+      <statement>
+        <p>
+           Suppose we replace every <m>\cos\theta</m> with <m>x</m> to get <m>2x^2-5x+3=0</m>. What are the factors of this equation?
```
AbbyANoble, line 732:
```suggestion
           Notice that this trigonometric equation looks similar to a quadratic equation. Let's replace every <m>\cos\theta</m> with <m>x</m> to get <m>2x^2-5x+3=0</m>. What are the factors of the left hand side of this equation?
```

AbbyANoble: Also, I think the equation was supposed to have a -3 at the end, not a +3.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+        <p>
+          <m>2\cos^2\theta-4=7\cos\theta</m>
+        </p>
+      </statement>
+      <answer>
+        <p>
+          <m>\theta=\frac{2\pi}{3}, \frac{4\pi}{3}</m>
+        </p>
+      </answer>
+    </task>
```
AbbyANoble, line 855: I think it might be good to change this to a sine one instead of cosine. They just solved $2\cos \theta +1 =0$ in part (a).

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+          <m>\theta=\frac{\pi}{3}, \pi, \frac{5\pi}{3}</m>
+        </p>
+      </answer>
+    </task>
+    </activity>
+
+    <remark>
+      <p>
+        In these last set of activities, let's look at another type of trigonometric equation you will see. These equations are multiple angle trigonometric equations. 
+      </p>
```
AbbyANoble, line 874:
```suggestion
      <p>
        In this last set of activities, let's look at trigonometric equations that involve multiple angles. 
      </p>
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+    <introduction>
+      <p>
+        Consider the equation <m>\sin(2\theta)=\frac{1}{2}</m>. 
+      </p>
+    </introduction>
+
+    <task>
+      <statement>
+        <p>
+          Although we are given <m>\sin(2\theta)=\frac{1}{2}</m>, let's replace <m>2\theta</m> with <m>\alpha</m> to help us think about what values of <m>\alpha</m> would satisfy <m>\sin\alpha=\frac{1}{2}</m>. What values of <m>\alpha</m> would satisfy the equation <m>\sin\alpha=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```
AbbyANoble, line 888:
```suggestion
          Notice that this equation involves  <m>\sin(2\theta)</m>, a multiple angle, rather that <m>\sin(2\theta)</m> like we have been dealing with. Let's replace <m>2\theta</m> with <m>\alpha</m> so it looks more like the equations we've been solving so far. What values of <m>\alpha</m> would satisfy the new equation <m>\sin\alpha=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+        <p>
+          <m>\theta=\frac{5\pi}{6}+2\pi(n)</m>
+        </p>
+      </answer>
+    </task>
+
+    <task>
+      <statement>
+        <p>
+           For <m>\sin\alpha=\frac{1}{2}</m>, we know the general solution is  <me>\alpha=\frac{\pi}{6}+2\pi(n)</me> and <me>\alpha=\frac{5\pi}{6}+2\pi(n)</me>. And because <m>\alpha=2\theta</m>, we can say <m>\sin(2\theta)=\frac{1}{2}</m>. Therefore, <me>2\theta=\frac{\pi}{6}+2\pi(n)</me> and <me>2\theta=\frac{5\pi}{6}+2\pi(n)</me>. Solve each general equation to find values of <m>\theta</m> that satisfies the equation <m>\sin(2\theta)=\frac{1}{2}</m>.
```
AbbyANoble, line 924:
```suggestion
           For <m>\sin\alpha=\frac{1}{2}</m>, we know the general solution is  <me>\alpha=\frac{\pi}{6}+2\pi(n)</me> and <me>\alpha=\frac{5\pi}{6}+2\pi(n)</me>. And because <m>\alpha=2\theta</m>, we can replace <m>\alpha</m> with  <m>2\theta</m> in those solutions. Therefore, <me>2\theta=\frac{\pi}{6}+2\pi(n)</me> and <me>2\theta=\frac{5\pi}{6}+2\pi(n)</me>. Solve each general equation to find values of <m>\theta</m> that satisfies the equation <m>\sin(2\theta)=\frac{1}{2}</m>.
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+      <hint>
+        <p>
+          Solve for <m>\theta</m>.
+        </p>
+      </hint>
+      <aside>
+        <p>
+        Recall that  <m>y=\sin(2\theta)</m> is a horizontal compression by a factor of <m>2</m> of the function  <m>y=\sin\theta</m>. On an interval of <m>2\pi</m>, we can graph two periods of  <m>y=\sin(2\theta)</m>, as opposed to one cycle of <m>y=\sin\theta</m>. This compression of the graph suggests there may be twice as many <m>x</m>-intercepts or solutions to  <m>\sin(2\theta)=\frac{1}{2}</m> compared to  <m>\sin\theta=\frac{1}{2}</m>.
+        </p>
+      </aside>
```
AbbyANoble, line 936: This aside isn't showing up in the preview for me.

AbbyANoble: Also, I might would make it an observation after the activity instead of an aside! It's informs why we're doing extra steps here.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+        <p>
+          <m>\theta=\frac{5\pi}{12}+\pi(n)</m>
+        </p>
+      </answer>
+    </task>
+
+    <task>
+      <statement>
+        <p>
+          Let's now test values of <m>n</m> to find all the values of <m>\theta</m>. Suppose <m>n=0</m>, what values of <m>\theta</m> would satisfy the equation <m>\sin(2\theta)=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```
AbbyANoble, line 950:
```suggestion
         We now need to find all values of <m>\theta</m> that satisfy our original equation  on the interval <m>[0, 2\pi)</m>. We can find these coterminal angles by substituting values of <m>n</m> in the general solution. We will start with <m>n=0</m>. What values of <m>\theta</m> result when $n=0$?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+        <p>
+          <m>\theta=\frac{5\pi}{12}+\pi(0)=\frac{5\pi}{12}</m>
+        </p>
+      </answer>
+    </task>
+
+    <task>
+      <statement>
+        <p>
+           Suppose <m>n=1</m>, what values of <m>\theta</m> would satisfy the equation <m>\sin(2\theta)=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```
AbbyANoble, line 966:
```suggestion
           Now let <m>n=1</m>. what values of <m>\theta</m> would satisfy the equation <m>\sin(2\theta)=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+        <p>
+          <m>\theta=\frac{5\pi}{12}+\pi(1)=\frac{17\pi}{12}</m>
+        </p>
+      </answer>
+    </task>
+
+    <task>
+      <statement>
+        <p>
+           Suppose <m>n=2</m>, what values of <m>\theta</m> would satisfy the equation <m>\sin(2\theta)=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```
AbbyANoble, line 982:
```suggestion
           Suppose <m>n=2</m>. What values of <m>\theta</m> would satisfy the equation <m>\sin(2\theta)=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+      <answer>
+      <p>
+          <m>\theta=\frac{\pi}{12}+\pi(2)=\frac{25\pi}{12}</m>, but this value is too big!
+        </p>
+        <p>
+          <m>\theta=\frac{5\pi}{12}+\pi(2)=\frac{29\pi}{12}</m>, but this value is too big!
+        </p>
+    </answer>
+    </task>
+
```
AbbyANoble, line 994: I think we need to check n=-1 too, and show the values are too small. Then some sort of summary that once we reach values outside of [0,2pi) in either direction, we can stop substituting values of n in that direction.

source/precalculus/source/08-TE/03.ptx
```diff
@@ -12,6 +12,1000 @@
...
+    <task>
+      <statement>
+        <p>
+          Look back at parts (d)-(f). What are all the solutions to <m>\sin(2\theta)=\frac{1}{2}</m>?
```
AbbyANoble, line 998:
```suggestion
          Look back at parts (d)-(f). What are all the solutions to <m>\sin(2\theta)=\frac{1}{2}</m> on the interval <m>[0, 2\pi)</m>?
```

source/precalculus/source/08-TE/03.ptx
tdegeorge: We don't require it, but I am okay to put it in.

tdegeorge:
@siwelwerd I believe I have addressed all the comments/suggestions...

siwelwerd (approved):

PR #665, opened 2025-03-17
ensure interesting AT3 scenarios

StevenClontz:
By making the codomain have odd dimension, we make sure kernel and image dimensions are different

### DIFF
--- a/source/linear-algebra/exercises/outcomes/AT/AT3/generator.sage
+++ b/source/linear-algebra/exercises/outcomes/AT/AT3/generator.sage
@@ -6,7 +6,6 @@ class Generator(BaseGenerator):
         # create a mxn standard matrix
         columns = 5
         rows = 4
-        maxrank = 4
         rank = choice([2,3])
         nullity = columns-rank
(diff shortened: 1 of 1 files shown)


siwelwerd:
Hmm, so it looks like this makes the rank-nullity equation always 2+3=5 or 3+2=5.  Wondering if students will pattern match and just always write down 3+2=5 for that part, and then be unhappy when they are told their explanation is insufficient.

StevenClontz:
What I didn't want is 2+2=4 (or any situation where rank=nullity).

Maybe these are the right scenarios (with a coin flip on whether the transformation is R4->R5 or R5->R4)

- rank+nullity=1+3=4=dim(Dom)
- rank+nullity=3+1=4=dim(Dom)
- rank+nullity=2+3=5=dim(Dom)
- rank+nullity=3+2=5=dim(Dom)

siwelwerd:
> What I didn't want is 2+2=4 (or any situation where rank=nullity).

Sure, that makes sense.

I think what you proposed will make a good enough mix and avoid equal rank and nullity.

siwelwerd (approved):

PR #666, opened 2025-03-17
Refactor PR7

### LINKED ISSUE #661: PR6 version 1 wrong graph
The graph shown in the solution looks something like -1/(x-2), as though the (x+5)^2 is missing.  When I graph the function in desmos it has a slant asymptote.

https://tbil.org/preview/precalculus/exercises/#/bank/PR6/1/

siwelwerd:
Closes #661 

Tweaked the plots (answers to Task 6) so that the aspect ratio isn't out of whack, which led to some graphs where important features were obscured.

Also refactored to remove some code duplication.

Example: 

Old
image

New
image

### DIFF
--- a/source/precalculus/exercises/outcomes/PR/PR7/generator.sage
+++ b/source/precalculus/exercises/outcomes/PR/PR7/generator.sage
@@ -110,7 +110,7 @@ class Generator(BaseGenerator):
         P=plot(f(x), (a,b), ymin=c,ymax=d,detect_poles=True,aspect_ratio=1,gridlines=True,ticks=[int((d-c)/10),int((d-c)/10)])
 
         #Undefined point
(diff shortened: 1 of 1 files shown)


siwelwerd:
The preview (and presumably others) action is broken this morning due to some upstream tinkering, so here's a screenshot of one of the graphs with the open dot

image

AbbyANoble:
> I don't think this is that easy or necessarily desirable. The prompt says to "Be sure to include all important points and asymptotes you've found.", so adding other points might imply that students should also have determined them to be important, rather than helpful in the plot. So I think it might just add confusion.

I guess I do deem those important! I have students plot points in each piece of the graph in addition to the intercepts and holes. Otherwise it's hard to know where to actually put your graph.

tdegeorge (approved):

PR #667, opened 2025-03-17
Update PF1 graphs

### LINKED ISSUE #533: Update labels on trig graphs

siwelwerd:
Closes #533

### DIFF
--- a/source/common/sagemath/library.sage
+++ b/source/common/sagemath/library.sage
@@ -270,7 +270,7 @@ class TBIL:
             
         kwds['ticks']=[custom_ticks,yticks]
         if yticks==1:
-            yticklabel = 1
+            yticklabel = SR(1)
         else:
             yticklabel = "latex"
         kwds['tick_formatter']=[custom_tick_labels,yticklabel]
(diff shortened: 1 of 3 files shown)


source/common/sagemath/library.sage
```diff
@@ -260,8 +268,12 @@ class TBIL:
         if 'thickness' not in kwds.keys():
             kwds['thickness']=3
             
-        kwds['ticks']=[custom_ticks,0.5]
-        kwds['tick_formatter']=[custom_tick_labels,"latex"]
+        kwds['ticks']=[custom_ticks,yticks]
+        if yticks==1:
+            yticklabel = 1
```
siwelwerd, line 273:
```suggestion
            yticklabel = SR(1)
```
See #669

siwelwerd:
This was superseded by #670

PR #670, opened 2025-03-20
Fix tick_formatter backwards compatibility

### LINKED ISSUE #669: Sage upgrade broke tick_formatters
Recent updates to the docker mean we have a (long needed) newer version of sage (now on 10.5) in our toolkit.  Unfortunately, this is not backwards compatible.  Some images are no longer building; the log shows this one is the current holdup:

https://github.com/TeamBasedInquiryLearning/library/blob/2d4d91477f2c355ebe7c0c7049ef59deb1ca4bc8/source/precalculus/source/meta/graphs.ptx#L174-L182

This actually builds fine in sagecell.sagemath.org, which runs 10.2.   However, now we are getting an error with the tick_formatter of 1; the error is that this is an integer, but Sage needs it to be an Expression.   So the fix is to replace `1` with `SR(1)`.

This likely shows up many many places--I think we use it any time we use a trig graph, for example, but possibly others.  I was just updating trig graphs to use the library in #667, so that may take care of a bunch of these.

siwelwerd:
Closes #669

### DIFF
--- a/source/common/sagemath/library.sage
+++ b/source/common/sagemath/library.sage
@@ -244,7 +244,7 @@ class TBIL:
         else:         
             xmin=0
             xmax=2*pi
-        if 'ticks' not in kwds.keys(): # or type(kwds['ticks']) is not sage.symbolic.expression.Expression:
+        if 'ticks' not in kwds.keys():
             delta=pi/4
             yticks=0.5
         elif type(kwds['ticks']) is sage.symbolic.expression.Expression:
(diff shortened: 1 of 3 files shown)


siwelwerd:
Oops, I branched this off of the branch from #669 to test that it fixed all of the things, so this has those commits here as well.  They are all needed to hopefully get the build working

StevenClontz (approved):
Thanks for tracking this down; I tried and failed to fix this myself earlier today.

Consider #671 and #672 for possible cleanups of the library function. Approving so you can merge after you do.

siwelwerd:
> Thanks for tracking this down; I tried and failed to fix this myself earlier today.

I'm sure there are other great new features, but PreTeXt 2.15.0 should actually tell us why sage images are failing :), which would make tracking down this kind of error easier.  I was going to try to test it and see if it broke anything this afternoon but ran out of time...
 
> Consider #671 and #672 for possible cleanups of the library function. Approving so you can merge after you do.

I don't quite understand #672 right now, so I merged in #671 so that hopefully the library will build overnight.  Feel free to convince me of the merits of #672 tomorrow.

PR #671, opened 2025-03-20
Tweak to PR670

StevenClontz:
Too involved to propose as suggestion, but I thought this might be easier to maintain.


siwelwerd (approved):
Yes, this is cleaner, thanks.

PR #672, opened 2025-03-20
Alternative tweak for #670

StevenClontz:


source/common/sagemath/library.sage
```diff
@@ -244,18 +244,17 @@ class TBIL:
...
-        elif type(kwds['ticks']) is list and len(kwds['ticks'])==2:
-            delta=kwds['ticks'][0]
-            yticks=kwds['ticks'][1]
         else:
-            delta=pi/4
-            yticks=0.5
+            try:
+                iter(kwds['ticks']) # raises TypeError if not iterable
+                delta=SR(kwds['ticks'][0])
+                yticks=SR(kwds['ticks'][1])
```
siwelwerd, line 254: I don't think this is quite what we want, lots of things like dictionaries and strings are iterables.  I guess a string shouldn't error out since you can index it, but it won't behave as expected.  A dictionary will cause an error.  So I'm inclined to stick with the simpler improvement implemented in #671

siwelwerd:
Closing (see comment above), reopen and tweak accordingly if you feel strongly that this is the way to go.

PR #676, opened 2025-03-21
fix site generation via updating requirements

### LINKED ISSUE #673: Website links missing
The homepage seems to no longer contain links to other pages along the top, and I get a 404 error from https://tbil.org/library/ , which I thought is the URL where the library was living.

On the plus side, I do see that the last commit built properly and there is a fresh copy of the books at e.g. https://tbil.org/preview/precalculus/instructor/

StevenClontz:
Closes #673

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
-pretext == 2.13.5
-pelican == 4.9.1
+pretext == 2.15.0
+pelican == 4.11.0
+markdown == 3.7.0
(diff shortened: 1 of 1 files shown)


siwelwerd:
The website looks good in the preview. Going to fire up a codespace and make sure the books compile still with the new PreTeXt--feeling a little trigger shy since the Sage upgrade broke things.

siwelwerd (approved):

PR #696, opened 2025-04-03
PF3 first draft

siwelwerd:
I'm not sure if we want more of the transformations here, but it was feeling a bit long as is, so posting this for feedback before writing any more.

### DIFF
--- a/source/precalculus/source/07-PF/03.ptx
+++ b/source/precalculus/source/07-PF/03.ptx
@@ -671,20 +671,20 @@ P=TBIL.trig_plot(sin(x),(x,-2*pi,2*pi),ymin=-pi,ymax=pi,ticks=[pi/4,1],aspect_ra
       </task>
       <task>
         <statement>
-          <p>Where does <m>\sin(x)</m> have maximal and minimal values?</p>
+          <p>Where does <m>\sin(x)</m> have local maximum and minimum values?</p>
         </statement>
         <answer>
-          <p>Maxima are at <m>\ldots,-\dfrac{3\pi}{2},\dfrac{\pi}{2},\dfrac{5\pi}{2},\ldots</m>.</p>
-          <p>Minima are at <m>\ldots,-\dfrac{\pi}{2},\dfrac{3\pi}{2},\dfrac{7\pi}{2},\ldots</m>.</p>
+          <p>Local maxima are at <m>\ldots,-\dfrac{3\pi}{2},\dfrac{\pi}{2},\dfrac{5\pi}{2},\ldots</m>.</p>
+          <p>Local minima are at <m>\ldots,-\dfrac{\pi}{2},\dfrac{3\pi}{2},\dfrac{7\pi}{2},\ldots</m>.</p>
         </answer>
       </task>
       <task>
         <statement>
-          <p>Where does <m>\csc(x)</m> have relative maximum and minimum values?</p>
+          <p>Where does <m>\csc(x)</m> have local maximum and minimum values?</p>
         </statement>
         <answer>
-          <p>Relative minima are at <m>\ldots,-\dfrac{3\pi}{2},\dfrac{\pi}{2},\dfrac{5\pi}{2},\ldots</m>.</p>
-          <p>Relative maxima are at <m>\ldots,-\dfrac{\pi}{2},\dfrac{3\pi}{2},\dfrac{7\pi}{2},\ldots</m>.</p>
+          <p>Local minima are at <m>\ldots,-\dfrac{3\pi}{2},\dfrac{\pi}{2},\dfrac{5\pi}{2},\ldots</m>.</p>
+          <p>Local maxima are at <m>\ldots,-\dfrac{\pi}{2},\dfrac{3\pi}{2},\dfrac{7\pi}{2},\ldots</m>.</p>
         </answer>
       </task>
       <task>
@@ -712,6 +712,7 @@ P=TBIL.trig_plot(sin(x),(x,-2*pi,2*pi),ymin=-pi,ymax=pi,ticks=[pi/4,1],aspect_ra
           Use your answers to the previous tasks to plot the graph of <m>y=\csc(x)</m>.
           </p>
         </statement>
+        <hint><p>It may be helpful to first sketch the graph of <m>\sin(x)</m>.</p></hint>
         <answer>
           <p> 
             <image>
--- a/source/precalculus/source/07-PF/03.ptx
+++ b/source/precalculus/source/07-PF/03.ptx
@@ -994,6 +994,343 @@ p
     </observation>
 
 
+  <activity xml:id="activity-cot-table">
+      <introduction>
+        <p> Consider the cotangent function, <m>f(x)=\cot(x)</m>.
+        </p> 
+      </introduction>
+
+      <task>
+        <statement>
+          <p>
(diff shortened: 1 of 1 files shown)


AbbyANoble:
Should we change the title of this section to "Graphing Additional Trigonometric Functions"? Seems like graphing is the goal, but that's not clear with the title.

source/precalculus/source/07-PF/03.ptx
```diff
@@ -11,6 +11,1404 @@
...
+              <li><m>2\pi</m></li>
+              <li><m>4\pi</m></li>
+            </ol>
+        </statement>
+        <answer>
+          <p>B.</p>
+        </answer>
+      </task>   
+
+    </activity>
```
AbbyANoble, line 315: I think it might be good to include an observation here at the end of the tangent activity that is like [Observation 7.3.4](https://c4407158.tbil.pages.dev/preview/precalculus/instructor/PF3#PF3-3-5) and the one for cosecant. Point out that the vertical asymptotes are where cos(x)=0 because that was the denominator. Show the graph for more than [0,2pi], etc.

AbbyANoble: Now that I've gotten farther, I see you summarize tangent and cotangent together in [Observation 7.3.12](https://c4407158.tbil.pages.dev/preview/precalculus/instructor/PF3#PF3-3-9).

AbbyANoble:
> I'm not sure if we want more of the transformations here, but it was feeling a bit long as is, so posting this for feedback before writing any more.

It is getting long. One thing that might help is maybe we do the long chart for secant. Then recognize its relationship to cosine in finding the vertical asymptotes and the maxima and minima. Then the next activity could be to graph cosecant trying to use those same ideas rather than all the point plotting? Kind of skip right to the graph a little quicker.

Similarly, graph tangent as is. Then a new activity asking where tangents asymptotes would be. And it's x-intercepts. And maybe a few points to establish that each piece decreases instead of increases. Plotting multiples of pi/4 isn't too bad since it is 1's and -1's.

source/precalculus/source/07-PF/03.ptx
```diff
@@ -11,6 +11,1404 @@
...
+p
+                    </sageplot></image>
+                </figure>
+            </sidebyside>
+    </observation>
+
+    <remark>
+    <p>The graphs of all six  <xref ref="trigonometric-functions-graphs" text="title"/> are listed in <xref ref="graphs-of-common-functions"/>.</p>
+    </remark>
+
```
AbbyANoble, line 1301: Between this remark and activity, it would help to make an observation that what we learned previously about transformations of graphs will still apply here.

source/precalculus/source/07-PF/03.ptx
```diff
@@ -11,6 +11,1404 @@
...
+    <activity>
+    <introduction>
+      <p>Consider the function <m>g(x)=\sec(x+\pi)</m>.</p>
+    </introduction>
+    <task>
```
AbbyANoble, line 1306: Before asking to spot the graph, I think asking how the graph differs from sec(x) first would be helpful. Choices like shifted left pi, shifted right pi, up pi, down pi, etc. 

Though if this is a question, we may want to change g(x) to sec(x+pi/2) perhaps since sec(x+pi) and sec(x-pi) are the same graph. (Wouldn't want to throw that in as a first question!)

source/precalculus/source/07-PF/03.ptx
```diff
@@ -11,6 +11,1404 @@
...
+    </task>
+    </activity>
+
+    <activity>
+    <introduction>
+    <p>Consider the function <m>h(x)=\tan(2x)</m></p>
+    </introduction>
+    <task>
+      <statement>
+        <p>Graph <m>h(x)=\tan(2x)</m>.</p>
```
AbbyANoble, line 1375: I feel like this is a hard first question and needs some scaffolding since the goal is to be able to graph this. I'd start by asking for the transformation(s) that are involved and what properties that changes. Some questions like:

1) How does tan(2x) differ from tan(x)? 
A Horizontal compression
B Horizontal stretch
C Vertical compression
D Vertical stretch

2) Which of the following will change be different for tan(2x) as compared for tan(x)? Select all that apply.
A location of vertical asymptotes
B period
C domain
D range

Then something to bring about that the "main" (not sure a word for this) cycle of tan(x) occurs between -pi/2 to pi/2 and the points (-pi/4, -1), (0,0), (pi/4,1). Dividing those by 2 (because of the 2x/horizontal compression) gets us a "main" cycles between -pi/4 and pi/4 with (-pi/8, -1), (0,0), (pi/8,1). Then we can sketch in copies of that cycle. 

I'm not sure this makes sense what I'm saying. But that's kind of how I go about drawing them. Think of the "main" cycle with its asymptotes and "anchor" points, adjust according to transformations, then copy/paste.

siwelwerd:
> It is getting long. One thing that might help is maybe we do the long chart for secant. Then recognize its relationship to cosine in finding the vertical asymptotes and the maxima and minima. Then the next activity could be to graph cosecant trying to use those same ideas rather than all the point plotting? Kind of skip right to the graph a little quicker.
> 
> Similarly, graph tangent as is. Then a new activity asking where tangents asymptotes would be. And it's x-intercepts. And maybe a few points to establish that each piece decreases instead of increases. Plotting multiples of pi/4 isn't too bad since it is 1's and -1's.

Since I will be copying and pasting, @AbbyANoble can you look at [Activity 7.3.6](https://b2ce8fa9.tbil.pages.dev/preview/precalculus/instructor/PF3#activity-csc-table) and let me know what you think before I rework the cotangent one similarly?  Do we need multiple choice on the first few tasks?

AbbyANoble:
> Since I will be copying and pasting, @AbbyANoble can you look at [Activity 7.3.6](https://b2ce8fa9.tbil.pages.dev/preview/precalculus/instructor/PF3#activity-csc-table) and let me know what you think before I rework the cotangent one similarly? Do we need multiple choice on the first few tasks?

I think it looks good! 

In (c) and (d), you switch from maximal/minimal to maximum/minimum. I'd have a preference for them to match and to go with the "mum" ones. I'd also include the word relative in both questions. Or actually local. Looking back in the functions section we used local, not relative.

In (g), a hint that it might help to get started by sketching in the sine curve.

source/precalculus/source/07-PF/03.ptx
```diff
@@ -11,6 +11,1404 @@
...
+                <cell><m>\tan(x)</m> (as a decimal)</cell>
+              </row>
+              <row bottom="minor">
+                <cell><m>0</m></cell>
+                <cell></cell>
+                <cell></cell>              
+              </row>
+              <row bottom="minor"> 
+                <cell><m>\dfrac{\pi}{6}</m></cell>
+                <cell> <m>\dfrac{1}{\sqrt{3}}</m> </cell>
```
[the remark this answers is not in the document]
siwelwerd: This is a good point but I think beyond the scope of this section, so I made an issue #730 to track this.  My inclination is to leave things as is here and add commentary earlier that sometimes we use both as convenient.

If the community consensus is different, we can come back and tweak these.

siwelwerd:
I incorporated all of the above suggestions (except the note about rationalizing denominators, which I documented in #730 ).  Also since it's a bit shorter I felt better about adding one more activity that combines two transformations.

AbbyANoble (approved):

PR #716, opened 2025-04-15
TR3 feedback

AbbyANoble:
TR3 feedback for @tdegeorge

### DIFF
--- a/source/precalculus/source/06-TR/03.ptx
+++ b/source/precalculus/source/06-TR/03.ptx
@@ -57,7 +57,7 @@
     <statement>
       <p>
         What is the value of <m>x</m> (i.e., the length of the third side)?
-        <ol marker= "A." cols="1">
+        <ol marker= "A." cols="3">
           <li> <m>5</m> inches </li>
           <li> <m>\sqrt{5}</m> inches </li>
           <li> <m>25</m> inches </li>
@@ -134,7 +134,7 @@
   <task>
     <statement>
       <p>
-        Draw a picture of this right triangle and label the sides (refer back to <xref ref="def-sides-of-right-triangles"/> to help you label the sides).
+        Draw a picture of this right triangle and label the parts of the triangle given.
       </p>
     </statement>
     <answer>
@@ -177,12 +177,6 @@
 </activity>
 
 <remark>
-  <p>
-  The Pythagorean Theorem can be very helpful in finding the third side of a right triangle as long as we know the length of two other sides. In <xref ref="intro-to-using-trig"/>, only one side and one angle were given. In this case, the Pythagorean Theorem is not enough to help us find another side of the right triangle. 
-  </p>
-</remark>
-
-<definition xml:id="def-sides-of-right-triangles">
   <statement>
     <p>
       When working with right triangles, it is often helpful to refer to specific angles and sides. One way this is done is by using letters, such as <m>A</m> and <m>a</m> to show that these are an angle-side pair because every angle has a side opposite the angle in a triangle. Note that the capital letter indicates the angle, and the lower case letter indicates the side. </p>
@@ -214,11 +208,7 @@
       </figure>
     </p>
   </statement>
-</definition>
-
-
-
-
+</remark>
 
 <definition xml:id="def-right-triangle-trigonometry">
   <statement>
@@ -252,11 +242,11 @@
   <task>
     <statement>
       <p>
-        Draw a right triangle and label it with angles <m>A</m>, <m>B</m>, and <m>C</m> as well as with the lengths given above. </p>
+        Draw a right triangle and label it with angles <m>A</m>, <m>B</m>, and <m>C</m>, with <m>c</m> being the hypotenuse of the triangle. </p>
     </statement>
     <answer>
       <p>
-        Students should draw a triangle so that <m>C</m> is the hypotenuse and each angle is across from the correct measure.
+        Students should draw a triangle so that <m>c</m> is the hypotenuse and each angle is across from the correct measure.
       </p>
     </answer>
   </task>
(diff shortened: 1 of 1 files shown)


AbbyANoble:
[Definition 6.3.7](https://tbil.org/preview/precalculus/instructor/TR3.html#def-sides-of-right-triangles) 
https://tbil.org/preview/precalculus/instructor/TR3.html#def-sides-of-right-triangles feels more like a remark than a definition. There's a lot going on!

Also, the image of the triangle in this one isn't scaling well. The letters get blurry if you zoom in some. Do we want to update this graphic?

source/precalculus/source/06-TR/03.ptx
```diff
@@ -62,6 +62,7 @@
           <li> <m>\sqrt{5}</m> inches </li>
           <li> <m>25</m> inches </li>
           <li> <m>16</m> inches </li></ol>
+          <li> <m>\sqrt{7}</m> inches </li>
```
AbbyANoble, line 65: This is what you get if you treat 4 as the hypotenuse.

source/precalculus/source/06-TR/03.ptx
```diff
@@ -203,7 +211,7 @@
 
 <remark>
   <p>
-  The Pythagorean Theorem can be very helpful in finding the third side of a right triangle as long as we know the length of two other sides. In <xref ref="intro-to-using-trig"/>, only one side and one angle were given. In this case, the Pythagorean Theorem is not enough to help us find another side of the right triangle (unless it is one of the triples!). 
+  The Pythagorean Theorem can be very helpful in finding the third side of a right triangle as long as we know the length of two other sides. In <xref ref="intro-to-using-trig"/>, only one side and one angle were given. In this case, the Pythagorean Theorem is not enough to help us find another side of the right triangle.
```
AbbyANoble, line 214: I'm not sure I'm following how knowing only one length and that it's a Pythgorean triple helps me find the other two? Like if one side is length 12, it could be a 5-12-13 or a 9-12-15.

source/precalculus/source/06-TR/03.ptx
```diff
@@ -201,9 +175,45 @@
...
+      <figure xml:id="labeling-sides-of-right-triangle">
+        <image source="Images/Defining-sides-of-a-triangle.png"/>
+        <caption> From the perspective of angle <m>A</m>, all sides of a right triangle can be labeled.</caption>
+      </figure>
+    </p>
+  </statement>
+</definition>
+
+
+
```
AbbyANoble, line 213: I think it might make more sense to put this definition after the activity where the pyth theorem wasn't enough. Through that activity, you realize you need more, so we define the opposite/adjacent idea and then the trig functions.

source/precalculus/source/06-TR/03.ptx
```diff
@@ -55,6 +62,7 @@
           <li> <m>\sqrt{5}</m> inches </li>
           <li> <m>25</m> inches </li>
           <li> <m>16</m> inches </li></ol>
+          <li> <m>\sqrt{7}</m> inches </li>
```
AbbyANoble, line 65:
```suggestion
          <li> <m>16</m> inches </li>
          <li> <m>\sqrt{7}</m> inches </li>
          </ol>
```

source/precalculus/source/06-TR/03.ptx
```diff
@@ -224,23 +236,98 @@
...
+      </p>
+    </statement>
+    <answer>
+      <p>
+        D
+      </p>
+    </answer>
+  </task>
+
+</activity>
```
AbbyANoble, line 317: Before we jump in to thinking which trig value to use to lead to solving a triangle, I thought it would be helpful to find trig values of an angle first.

source/precalculus/source/06-TR/03.ptx
```diff
@@ -224,23 +236,98 @@
...
+
   <activity xml:id="determine-which-trig-function">
     <introduction>
       <p>
-        For each triangle given, determine which trigonometric function would be the most helpful in determining the length of the side of a triangle. Be sure to draw a picture of the triangle to help you determine the relationship between the given angle and sides.
+        For each triangle given, determine which trigonometric function would be the most helpful in determining the length of the side of a triangle. Be sure to draw a picture of the triangle to help you determine the relationship between the given angle and sides. In each case assume angle <m>C</m> is the right angle.
```
AbbyANoble, line 322: Would it help to walk through and solve one of these before we move on to the application one?

AbbyANoble:
Regarding [Activity 6.3.13](https://tbil.org/preview/precalculus/instructor/TR3.html#applying-trig):

- I feel like it's unclear whether we are looking at the Eiffel Tower from a horizontal distance or hypotenuse type distance of 500 meters. 
- Throwing an alpha in felt strange as it's the first time we've seen it?
- In part (a), it says use x to represent the missing side. But we only know one side, so there are two missing sides.
- I feel like it's a big jump to solve this, especially given that we are approximating in the calculator. I don't know that we've built up that each angle has its own specific trig values that the calculator knows regardless of the triangle I'm looking at. It feels mystical here.


To help with that last one, maybe before this we need to explore trig values more before we use them to solve things. Like maybe show two 30-60-90 triangles of different sizes. Then find trig values of both 30 degree angles and notice they are the same. So the trig value doesn't depend on the size of the triangle, but the ratio of the side lengths.

AbbyANoble:
[Remark 6.3.18](https://tbil.org/preview/precalculus/instructor/TR3.html#TR3-3-17) 
and the activities that depend on it feel out of place here. Seems like these type of problems should wait until we've discussed the inverse trig functions in PF4? E

tdegeorge:
> https://tbil.org/preview/precalculus/instructor/TR3.html#def-sides-of-right-triangles

I'm okay with changing it to a remark. I just wanted something in there that reminds students how to label the different sides as I've seen them get confused on that.

source/precalculus/source/06-TR/03.ptx
tdegeorge: Good point. I was trying to convey that that are some triples that are more common and that if we know them, we can bypass some of the "work"

source/precalculus/source/06-TR/03.ptx
tdegeorge: I like that.

source/precalculus/source/06-TR/03.ptx
tdegeorge: I didn't put the application in after solving it because it only required them to find the height of the Eiffel Tower. I would prefer that we put it right before the solving

tdegeorge:
> Regarding [Activity 6.3.13](https://tbil.org/preview/precalculus/instructor/TR3.html#applying-trig):
> 
> * I feel like it's unclear whether we are looking at the Eiffel Tower from a horizontal distance or hypotenuse type distance of 500 meters.

Shouuld we add in something like "an observer is standing 500 meters from the Eiffel Tower..."?

> * Throwing an alpha in felt strange as it's the first time we've seen it?

We can use theta if you want. I am always conflicted on if we should be representing more than one symbol. Just like when we solve for a variable, should we always use x?

> * In part (a), it says use x to represent the missing side. But we only know one side, so there are two missing sides.

Maybe say, "Use x to refer to the height of the Eiffel Tower"?

> * I feel like it's a big jump to solve this, especially given that we are approximating in the calculator. I don't know that we've built up that each angle has its own specific trig values that the calculator knows regardless of the triangle I'm looking at. It feels mystical here.

I'm not sure I understand? Should we add in another scaffolded question before the set up part?
> 
> To help with that last one, maybe before this we need to explore trig values more before we use them to solve things. Like maybe show two 30-60-90 triangles of different sizes. Then find trig values of both 30 degree angles and notice they are the same. So the trig value doesn't depend on the size of the triangle, but the ratio of the side lengths.

tdegeorge:
> [Remark 6.3.18](https://tbil.org/preview/precalculus/instructor/TR3.html#TR3-3-17) and the activities that depend on it feel out of place here. Seems like these type of problems should wait until we've discussed the inverse trig functions in PF4?
> 
> Maybe this section finds trig values but doesn't solve triangles. Then maybe between TE3 and TE4 (law of sines/cosines) we do a section of solving right triangles and applications? That kinds of makes sense right before a section of solving non-right triangles.

I'm pretty sure we agreed to put this in the other section. I can't, however, seem to find this branch to make the changes....

AbbyANoble:
> > Regarding [Activity 6.3.13](https://tbil.org/preview/precalculus/instructor/TR3.html#applying-trig):
> > 
> > * I feel like it's unclear whether we are looking at the Eiffel Tower from a horizontal distance or hypotenuse type distance of 500 meters.
> 
> Shouuld we add in something like "an observer is standing 500 meters from the Eiffel Tower..."?
> 

Yeah. Maybe "Tonya is standing 500 meters from the base of the Eiffel Tower. She looks at the top of the tower at an angle of 31 degrees."  That also gets rid of the variable. ( I also just noticed you say alpha in the question, but use theta in the answer.)

> > * Throwing an alpha in felt strange as it's the first time we've seen it?
> 
> We can use theta if you want. I am always conflicted on if we should be representing more than one symbol. Just like when we solve for a variable, should we always use x?
> 
> > * In part (a), it says use x to represent the missing side. But we only know one side, so there are two missing sides.
> 
> Maybe say, "Use x to refer to the height of the Eiffel Tower"?
> 

Yeah, I think that is clearer.

> > * I feel like it's a big jump to solve this, especially given that we are approximating in the calculator. I don't know that we've built up that each angle has its own specific trig values that the calculator knows regardless of the triangle I'm looking at. It feels mystical here.
> 
> I'm not sure I understand? Should we add in another scaffolded question before the set up part?
> 

Maybe have them solve for x between setting up the equation and giving the answer from the calculator?

> > To help with that last one, maybe before this we need to explore trig values more before we use them to solve things. Like maybe show two 30-60-90 triangles of different sizes. Then find trig values of both 30 degree angles and notice they are the same. So the trig value doesn't depend on the size of the triangle, but the ratio of the side lengths.

tdegeorge:
@siwelwerd I believe I fixed all the issues with this section. When you get a chance, can you double check and then we need to move the finding an angle part to a different section (starting with 6.3.19) - I can't find where we said we wanted to have a solving part separate, but I remember us saying we wanted to shift some of the sections around...

siwelwerd:
> I can't find where we said we wanted to have a solving part separate, but I remember us saying we wanted to shift some of the sections around...

I made a note in #718

siwelwerd (approved):

PR #720, opened 2025-04-16
GT1 interactives and scaffolded exploration tweaks

StevenClontz:


StevenClontz:
Will teach with this in my next class then mark as ready for review if no further work is needed.

StevenClontz:
Need to stop fiddling with this - there's plenty enough to review for now...

PR #722, opened 2025-04-17
update GT1 sample to match checkit

StevenClontz:


StevenClontz:
Abusing my powers to merge so this is live for class tomorrow 😶

PR #723, opened 2025-04-17
update codespace config

StevenClontz:
This should get us back to a working codespace. As a bonus, this new model boots up fairly instantly - several dependencies are installed in the background, but an author can start writing while they install.

Can test this by typing a comma `,` while viewing this PR - it should try to start up a codespace with this configuration.


siwelwerd:
Seems to boot up okay, but codespace takes a while (still waiting) in a way that is not transparent to the user (unless they are willing to look at the creation logs)

StevenClontz:
I tried to make this clear in our welcome.md - what do you think?

siwelwerd (approved):

PR #727, opened 2025-04-21
EQ5 Checkit remove degenerate case

### LINKED ISSUE #725: EQ5 CheckIt
The answer to the second part of task 1 recommends completing the square which seems off. I'm assuming something in the way these were generated made this one a little different than intended. 

Image

siwelwerd:
Closes #725 

I don't know if this will show up in the previews, but the completing the square tasks were all of the form $x^2+2ax=b$.  This fix removes the possibility of $a$ being zero, which would result in $x^2=b$, which is probably best solved by the square root property rather than completing the square.

### DIFF
--- a/source/precalculus/exercises/outcomes/EQ/EQ5/generator.sage
+++ b/source/precalculus/exercises/outcomes/EQ/EQ5/generator.sage
@@ -27,7 +27,7 @@ class Generator(BaseGenerator):
 
     #Completing the square
 
-    a=choice([-7..7])
+    a=choice([-7..-1,1..7])
     b=choice([i for i in [-1*a^2+5,..,10] if not (i+a^2).is_square()])
     r1=-1*a+sqrt(b+a^2)
     r2=-1*a-sqrt(b+a^2)
--- a/source/precalculus/exercises/outcomes/EQ/EQ5/generator.sage
+++ b/source/precalculus/exercises/outcomes/EQ/EQ5/generator.sage
@@ -34,10 +34,12 @@ class Generator(BaseGenerator):
     equations.append( { "equation":  x^2+2*a*x== b , "method":"completing the square", "roots": f"{latex(r1)}\\text{{ and }}{latex(r2)} "})
 
     #Quadratic Equation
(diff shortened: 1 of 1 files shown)


AbbyANoble:
Just going through the 20 samples, I didn't see it happening again so that's good!

I did see this, though. Notice it's suggesting quadratic formula for 4x^2=8. Probably a similar fix needed there!

Screenshot 2025-04-21 at 4 38 56 PM

siwelwerd:
Hmm okay, marking as draft and will check that out too

source/precalculus/exercises/outcomes/EQ/EQ5/generator.sage
```diff
@@ -1,7 +1,7 @@
 load("../../../source/common/sagemath/library.sage")
 
 class Generator(BaseGenerator):
-  irrational_dict = TBIL.small_irrationals(rational_part=[-8..1,1..8],full_list=True)
+  irrational_dict = TBIL.small_irrationals(rational_part=[-8..-1,1..8],full_list=True)
```
siwelwerd, line 4: Missing negative allowed the rational part to be 0 in some cases, resulting in the sum of the roots being zero, and thus the coefficient of $x$ is zero.

siwelwerd:
Found and fixed that @AbbyANoble , let me know if you notice any other weirdnesses!  Otherwise this is ready.

AbbyANoble:
So this is more pedagogical than technical, but how are we deciding between completing the square and quadratic formula? I _think_ the completing the square ones have an even $b$ value maybe? And have $a=1$. But these seem about the same to me in the screenshot below. (Though, honestly, I'm probably rarely going to choose to complete the square.) Some of the other samples that lead to quadratic formula have $a\neq1$, which, agreed. I'm pretty much defaulting to that unless I can see how to factor it fairly fast.

Screenshot 2025-04-21 at 8 26 59 PM

AbbyANoble (approved):
Looks okay now. Not seeing any issues in the preview!

PR #735, opened 2025-04-30
MX4: use tech to verify product matrix applies row ops

StevenClontz:


jkostiuk (approved):
nice, i like this a lot.

PR #740, opened 2025-05-12
Move hint

siwelwerd:
There was a `<hint>` inside of the `<introduction>` of an `<activity>`, which built fine (but did not show up) in the HTML, but caused an exercise component mislabeled error in the PDF.

I tried moving it as a sibling of the `<introduction>`, but in this case both the HTML and PDF built fine but quietly did not display the hint.

Workaround here is to add it as a child of the first task, which should fix the build.

### DIFF
--- a/source/precalculus/source/08-TE/04.ptx
+++ b/source/precalculus/source/08-TE/04.ptx
@@ -1014,12 +1014,8 @@ p
       <p>
         State whether the Law of Sines or the Law of Cosines is the best choice to use to determine the indicated angle/side.
       </p>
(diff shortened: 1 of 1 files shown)


StevenClontz (approved):

StevenClontz:
I think `<hint/>`s are structured as `<answer/>` et al, so it doesn't surprise me they only work where an answer would work.

siwelwerd:
I'm less bothered that they don't work than that they behave inconsistently and don't throw errors when they don't work.

PR #741, opened 2025-05-12
Make welcome show up formatted

StevenClontz:


StevenClontz:
😤 not sure why this works at https://github.com/pi-base/data/blob/main/.devcontainer.json but not here...

siwelwerd:
Interesting, it doesn't render the markdown when I boot up a codespace, but if you close the file and reopen it it does.  So it seems it's opening the file before changing the setting, not sure how you convince it to do the other order though.

StevenClontz:
While it doesn't work today, I believe it's *supposed* to work, so I'm inclined to merge this and hope one day they fix it.

siwelwerd (approved):
It doesn't seem to break anything, so no harm merging it.

PR #749, opened 2025-05-19
TR2 Checkit

siwelwerd:
TR2 Checkit

I needed an update one of the common library functions to typeset the angles nicely, so that's why the TR1 branch is merged in here. Hopefully I did that right.


tdegeorge (approved):
From what I can see, everything looks good....I went into the preview mode and looked at multiple versions.

PR #750, opened 2025-05-20
Addressed Issue TR4 feedback #726

### LINKED ISSUE #726: TR4 feedback
TR4 feedback for author @tdegeorge.
 
Trying out issues since doing it as a pull request was kind of a pain last time! Linking locations and putting suggestions **in bold**. I'm also putting suggestions in separate comments too so we can thumbs up/thumbs down/reply to each one. We will see how this goes!

tdegeorge:
Closes #726

### DIFF
--- a/source/precalculus/source/06-TR/04.ptx
+++ b/source/precalculus/source/06-TR/04.ptx
@@ -641,46 +641,46 @@
             </p>
           </answer>
         </task>
+
+        <task>
+          <statement>
+            <p>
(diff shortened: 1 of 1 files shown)


tdegeorge:
@siwelwerd I believe I addressed all Abby's comments. Codechat, however, kept saying there was a build error....can you check to see if there is something else that is not right? Thanks!

siwelwerd:
When you added those extra 3 tasks, looks like they got inserted below the `</activity>`, hence the error.  Just pushed a commit to fix that.

siwelwerd (approved):

Issue #714, opened 2025-04-14
"Typo" in GT1


jkostiuk:
In observation 5.1.12, the labelling of v,w are opposite in the visual than they are in algebraic calculation
<img width="823" alt="Image" src="https://github.com/user-attachments/assets/3e743364-7df8-465a-ac49-646fb73f4b3b" />

PR #762, opened 2025-05-21
addresses issue "Typo" in GT1 #714

megancl13:
Fixed the chain of equalities at the bottom of the observation to match the notation from the figure.


siwelwerd:
Thanks for this contribution Megan!

Issue #761, opened 2025-05-20
GT5 Sample Problem


siwelwerd:
https://github.com/TeamBasedInquiryLearning/library/blob/13098ccb5296a039bf08352ba877c3a9c969d92f/source/linear-algebra/source/05-GT/05.ptx#L191-L235

This has `<task>`s within an `<introduction>`, which seems wrong, and displays weird in the PDF.  This should be corrected, as well as factored out into a separate file and included in the Appendix like all of the other Sample Problems.

siwelwerd:
This was fixed by #769

PR #766, opened 2025-05-21
Addressing issue #761

jkostiuk:
moved sample solution to separate file, corrected formatting issues


siwelwerd:
This is set to merge into your branch `jordans-workspace`, should this instead be merged into main?

jkostiuk:
oh good catch---i made the same mistake with megan earlier lol

PR #767, opened 2025-05-21
AT1 Sample Solution

### LINKED ISSUE #685: AT1 Sample Solutions should have a counter example
In the sample solution for AT1 part (ii), where students are proving something is not a vector space, the sample solution computes the LHS and the RHS. It then looks at the general experessions and states that they're unequal rather than providing a counterexample. 

I think adding a counterexample is important here---how do we know these expressions are *actually* different and not just different forms of the same expression.

megancl13:
Changed the sample solution to include calculation of explicit counter-examples (for addition and scalar multiplication) instead of general vectors and scalar. 

Closes #685

### DIFF
--- a/source/linear-algebra/source/03-AT/samples/01.ptx
+++ b/source/linear-algebra/source/03-AT/samples/01.ptx
@@ -41,41 +41,40 @@
                that <m>S</m> fails to preserve scalar multiplication. 
             </p>
             <p>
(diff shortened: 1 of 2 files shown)


siwelwerd (approved):

StevenClontz:
Thanks for the contribution @megancl13!

PR #778, opened 2025-05-21
Update installSage.sh

StevenClontz:
This would match the behavior of our github action: 

https://github.com/TeamBasedInquiryLearning/library/blob/dd9a0a9cb7f9c508b907c671f1a6437af61f2328/.github/workflows/pretext-cli.yml#L26

Closes #777


siwelwerd:
This supercedes #777?

PR #776, opened 2025-05-21
add GT5 checkit

StevenClontz:
Closes #770

PR #785, opened 2025-05-22
Implementing change of basis outcome in MX

### LINKED ISSUE #780: Reorganizing/Merging of Some Linear Algebra Lessons (MX2, MX3, and change of basis/GT5)
I'd like to pitch a suggestion for re-organizing some content in Linear Algebra--original conversation started in #776 

Context: [Section GT5 of the activity book is change-of-basis](https://tbil.org/preview/linear-algebra/GT5.html). The file is in the main project, but I don't believe it's published as part of the main edition yet. My instructors and I taught with this in Fall and Spring and the lesson itself went well for us. Now that we're thinking about including this in the main edition, we want to take a step back and make sure the lesson is in the right place and that the other lessons around it flow. 

My specific suggestions for reorganizing: 

1. merge MX3 into MX2. Given the way we defined the inverse and inverse matrix, I think it follows naturally in that lesson that we can use the inverse matrix to solve a system of equations. then, in the MX2 checkit, we can have task 1 be "explain why this matrix isn't invertible" task 2 can be your reworked "find the inverse" and then we can add a short task 3 that's "use your inverse to solve the system". While this makes MX2 a little longer, the ideas are well-connected: we defined the inverse matrix/trafo to be "the unique solution to this equation" so it would make thematic sense for the lesson to end with "now we can easily compute this unique solution".
2. Then, GT5 becomes a new MX3.
3. If we do this, then I think we can add 2-3 activities to GT4 to do some version of diagonalization (i.e., after computing a basis of eigenvectors, use that to diagonalize a matrix).

What does the community think? (Note that change-of-basis cannot happen earlier than this suggestion without significant edits because it requires the matrix inverse).

Tagging some specific people, but all comments welcome here! 
@siwelwerd @StevenClontz @jford1906 @jbunn3 @megancl13

StevenClontz:
Closes #776 
Closes #780


jkostiuk:
@StevenClontz I think I'm all set here. I can't ask for your review because you're the author. If you or @siwelwerd could take a look, I'd be happy to build any additional changes (though i don't know how to resolve conflicts with the requirements.txt). 

Will address modifications to GT4 in a separate PR.

source/linear-algebra/source/04-MX/samples/02.ptx
```diff
@@ -4,139 +4,86 @@
...
+              <p>  Its inverse is <m>\left[\begin{array}{cccc} 4 &amp; -7 &amp; 6 &amp; -4 \\ -3 &amp; 9 &amp; -7 &amp; 4 \\ -1 &amp; 1 &amp; -1 &amp; 1 \\ -1 &amp; 2 &amp; -2 &amp; 1 \end{array}\right]</m>.   </p>
+            </solution>
+          </task>
+          <task>
+            <statement>
+              <p> If the matrix is invertible, explain how to find its inverse using techniques that don't require technology. (You may use technology to perform those techniques, as long as they *could* be done without using technology.) </p>
+            </statement>
+            <solution>
+              <p>  To calculate the inverse, we solve the vector equations <m>D\vec{x}=\vec{e}_i</m> for <m>i=1,2,3,4</m>. 
+            These solution vectors then form the columns of the inverse.   </p>
```
StevenClontz, line 36: So this would not be explicility copy-pasted into every solution, we will rewrite this + checkit to ask for a specific column.

source/linear-algebra/source/04-MX/03.ptx
```diff
@@ -11,325 +11,180 @@
 
 <subsection>
     <title>Warm Up</title>
```
StevenClontz, line 13: need to add warm-up

jkostiuk (approved):

Issue #743, opened 2025-05-13
TE5 (now TE6) feedback


AbbyANoble:
Feedback for TE5 for author @tdegeorge

AbbyANoble:
[Activity 8.5.2(b)](https://tbil.org/preview/precalculus/instructor/TE5.html#right-triangles-1-3) 

I think this would be hard to know where to draw without knowing what an angle of descent it. Maybe define the angle of descent in the question, then ask them to draw it in?

AbbyANoble:
[Definition 8.5.4](https://tbil.org/preview/precalculus/instructor/TE5.html#def-angle-of-depression-and-elevation) 

Checking to see where this picture is from and if it is okay to use!

AbbyANoble:
[Activity 8.5.6](https://tbil.org/preview/precalculus/instructor/TE5.html#right-triangles-2) 

The person's eyes are (presumably) not on the ground. Should be adjust to account for that?

AbbyANoble:
[Activity 8.5.7(a)](https://tbil.org/preview/precalculus/instructor/TE5.html#right-triangles-practice-2) 

Missing the word degrees after 70.

AbbyANoble:
[Activity 8.5.7](https://tbil.org/preview/precalculus/instructor/TE5.html#right-triangles-practice) 

All of these used either sine or tangent. Didn't know if we wanted to throw a cosine in too.

AbbyANoble:
[Activity 8.5.10(a)](https://tbil.org/preview/precalculus/instructor/TE5.html#law-of-cosines-and-sines-example-2) 

I'm not sure it's clear automatically that it's a non-right triangle. I think they'd need to verify with the pythagorean theorem or just assume it's not since LOC still would work.

And it might help for the hint to say use the letters C, J, and T rather than A, B, and C.

AbbyANoble:
[Activity 8.5.12(a)](https://tbil.org/preview/precalculus/instructor/TE5.html#periodic-example-1-2) 

Missing the slash in front of the pi for display.
<img width="618" alt="Image" src="https://github.com/user-attachments/assets/57929c04-6d21-4408-bdba-4fc69fce714d" />

AbbyANoble:
[Activity 8.5.13(b)](https://tbil.org/preview/precalculus/instructor/TE5.html#periodic-example-2-3) 

[...]

PR #790, opened 2025-05-23
Addresses TE5 (now TE6) feedback #743

tdegeorge:
#743


tdegeorge:
@siwelwerd The only thing I am not sure about is the image used for the angle of elevation and angle of depression. I got the image from here: https://www.teachoo.com/7155/1146/Angle-of-Depression-from-point-A-to-point-B-is-same-as-Angle-of-Elevat/category/Questions-easy-to-difficult/

siwelwerd:
How's this?  I don't think it will show in the preview.

image

tdegeorge:
> How's this? I don't think it will show in the preview.
> 
> image

That's perfect!

siwelwerd (approved):

PR #795, opened 2025-05-24
Less clunky authoring in Codespace

StevenClontz:
Implements two things:

- A new `authoring-preview` target that builds *everything*, so we can have a single `codechat-config.yaml` and no longer require shuffling files for previewing different books.
- Support to automatically zip up generated assets as `cache.zip` when building previews, then pull them down for editing, so there's no "first-time" asset generation friction.


StevenClontz:
Once https://tbil.org/static/cache.zip exists, I'll set up previews to pull the cache as well so that should *also* drastically speed up.

StevenClontz:
When reviewing, don't freak out that there are 100+ changed files. Everything in `assets` is just a bunch of moved files, not changed ones

PR #810, opened 2025-05-28
TR4 Checkit

siwelwerd:


tdegeorge:
@siwelwerd 

Can we also make the answers to these as one fraction? For example, sin (pi/3) = sqrt3/2 (but not 1/2*sqrt3).

Also, some of the fractions look really small. For example, the answer to sin(pi/6) is 1/2 and the 1/2 is much smaller in size the the font of the other part

tdegeorge (approved):

PR #816, opened 2025-06-16
PF2 Checkit

siwelwerd:


tdegeorge:
@siwelwerd Oh I see it now - it is PF3 in our outline....

but I still can't remember what we said about domain and range. Should we have them state it?

source/precalculus/exercises/bank.xml
```diff
@@ -374,6 +374,8 @@
                Determine the basic properties of the graphs of sine and cosine, including amplitude, period, and phase shift.
             </description>
         </outcome>
+    -->
+            
       <!--
```
siwelwerd, line 379:
```suggestion
```

Remove merge artifact

siwelwerd:
@tdegeorge I added tasks for the domain and range, and fixed the graphs

siwelwerd:
Oops.  Tried to fix a typo at the last second and also added in a new one.  Should be fixed now @tdegeorge

tdegeorge (approved):

PR #819, opened 2025-06-18
TE2 Checkit

siwelwerd:


tdegeorge (approved):
I went through and did a few of these to make sure they work. I think it looks good...the only thing is the typesetting stuff like the other check it (with sin^2). It looks like, though that there is something else blocking the merge.

PR #820, opened 2025-06-18
TE3 Checkit

siwelwerd:


tdegeorge:
@siwelwerd 

For some versions in Task 2, it looks like the theta is being squared rather than the trig function. For example, sin(theta)^2-1=0....shouldn't it be sin^2 (theta)-1=0?

PR #821, opened 2025-06-18
TE4 Checkit

siwelwerd:
This section was added, so there wasn't a sample problem for this outcome.  I modeled it after [Activity 8.4.5](https://tbil.org/preview/precalculus/TE4.html#solving-right-triangles)

### DIFF
--- a/source/precalculus/exercises/outcomes/TE/TE4/generator.sage
+++ b/source/precalculus/exercises/outcomes/TE/TE4/generator.sage
@@ -8,14 +8,18 @@ class Generator(BaseGenerator):
             A=round(arctan(a/b)*180/pi,1)
             B=round(arctan(b/a)*180/pi,1)
             C=90
+            sides_rounding="integer"
+            angles_rounding="tenth of a degree"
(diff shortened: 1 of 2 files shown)


tdegeorge:
@siwelwerd I'm not sure what is happening, but I am not getting the same answers for some of the versions (when given an angle and two sides)....Screenshot 2025-06-19 at 9 02 28 PM

Can we also specify what place to round to? For example, if the problem gives a decimal like 4.5 (for a given side), could we ask to have the answers rounded to the same place value? I think it is okay to have all the degree measures be rounded to the nearest degree.

siwelwerd:
> @siwelwerd I'm not sure what is happening

I was mixing up the two angles because trig is hard 🤦🏻‍♂️

> Can we also specify what place to round to? For example, if the problem gives a decimal like 4.5 (for a given side), could we ask to have the answers rounded to the same place value? I think it is okay to have all the degree measures be rounded to the nearest degree.

Added this in too.

tdegeorge (approved):

PR #832, opened 2025-06-26
upgrade ptx and docker container

StevenClontz:


siwelwerd:
> Not sure if that is fixable, or should just be documented for the stray user who wants to go that route.

Going ahead and approving since the Codespaces is fine, but advise if you want to try and fix this, or go ahead and merge it and we'll document the behavior

PR #847, opened 2025-07-23
special preview targets

StevenClontz:


StevenClontz:
https://library.tbil.org/2025/calculus/print/TBIL-Calculus.pdf

https://library.tbil.org/2025/calculus/print/TBIL-Calculus-Instructor.pdf

https://library.tbil.org/2025/calculus/print/TBIL-Calculus-Slides.pdf

I think we're good?

siwelwerd:
> I think we're good?

Nope, Calculus is still crapping out after 18 pages (I think when it gets to the first QR code).  Precalc student PDF looks good, but instructor PDF doesn't.

StevenClontz:
I think it's a caching issue: they look fine when I download them now.

But I need to check precal instructor PDF.



Steven Clontz
https://clontz.org - ***@***.***

siwelwerd:
I think the caching error has resolved, but the precalc instructor still wasn't working.  It wouldn't compile locally, but now it does. I guess there was an error in math mode, that MathJax would report as an error, but PreTeXt doesn't check that things in `<m>` tags are valid.  Thus building HTML reported success, but LaTeX failed.

Latest commit should fix this.

StevenClontz:
Building and deploying it now. If it looks good, I'll merge this so the website launches with the 2025 edition links.

PR #851, opened 2025-08-04
Remove notice of new website

siwelwerd:
Our website is no longer new, I think we can remove this notice.


StevenClontz (approved):
I agree. Instead I need to update the old site to redirect folks here.

PR #856, opened 2025-08-19
LE2: refactor RREF discovery to align with @davidaustinm #teamforwardsubstitution

### LINKED ISSUE #854: Gauss-Jordan elimination
I'm going to quibble about your presentation of the Gauss-Jordan elimination algorithm, in particular, [Activity 1.2.14](https://library.tbil.org/2025/linear-algebra/LE2.html#LE2-4-14) 
https://library.tbil.org/2025/linear-algebra/LE2.html#LE2-4-14 and the [video](https://www.youtube.com/watch?v=6iGMPpD9Mf8&list=PLwXCBkIf7xBMo3zMnD7WVt39rANLlSdmj&index=2).  

Your algorithm focuses on the pivot columns and produces 0's below *and above* the pivot position before moving on to the next column.  The more traditional algorithm produces 0's below the pivot position only and then goes to the next column, which breaks the algorithm into two phases:  forward substitution and back substitution.  I see this as preferable for the following reasons:
1. It reduces flops and is how computers implement the algorithm.  I realize that you're doing this "by hand" but a "flop" in this case represents student effort and the potential for mistakes (and this type of mental arithmetic already leads to lots of mistakes).
2. The upper triangular form obtained from forward substitution with less effort reveals a great deal about the matrix, namely its rank, a basis for the column space, the determinant, and more.  We often don't care about the full RREF.
3. Your implementation will not produce an LU factorization whereas LU is a natural byproduct of the traditional algorithm.
4. Pedagogically, this also allows for better scaffolding and motivation.  Clearly, a diagonal system is the easiest.  Starting with an upper triangular system, we can work right to left to get a diagonal system so it turns out that's pretty easy too.  So how can we get a general system into an upper triangular form?  Implement the forward substitution pass.

#teamforwardsubstitution

StevenClontz:
Closes #854

### DIFF
--- a/source/linear-algebra/source/01-LE/02.ptx
+++ b/source/linear-algebra/source/01-LE/02.ptx
@@ -906,61 +906,15 @@ This will keep you from undoing your progress towards an RREF matrix.
     </p>
 </observation>
 
-<activity estimated-time='10'>
-    <statement>
-        <p>
-Complete the following RREF calculation (multiple row operations may be needed
-for certain steps):
-<me>
-    A=
-    \left[\begin{array}{cccc}2 &amp; 3 &amp; 2 &amp; 3 \\ -2 &amp; 1 &amp; 6 &amp; 1 \\ -1 &amp; -3 &amp; -4 &amp; 1 \end{array}\right]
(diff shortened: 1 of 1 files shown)


StevenClontz:
I think I like this well enough. Though I have to comment that I'm covering this topic irregularly: because I'm out of town next week, I've pushed all fluency building there with the substitutes, and just did the scaffolded exploration with my students this week.

Tagging @jkostiuk for friendly criticism or suggestions.

PR #895, opened 2025-11-14
MX2 Fixes

StevenClontz:
Fix MX2 so the non-invertible task doesn't have an incomplete template.

Also shuffles the order to prompt the student to compute a column by hand first, and then verify with technology.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "githubPullRequests.ignoredPullRequestBranches": [
-        "main"
-    ]
-}


source/linear-algebra/exercises/outcomes/MX/MX2/generator.sage
```diff
@@ -43,6 +43,7 @@ class Generator(BaseGenerator):
             "invertible": False,
             "label": labels[1],
             "vector_eq": TBIL.VectorEquation(m),
+            "ordinal": ordinal_string,
```
StevenClontz: This will fix the non-invertible question so it doesn't ask for a [blank] column of the inverse.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "githubPullRequests.ignoredPullRequestBranches": [
+        "main"
+    ]
+}
```
siwelwerd: Should `.vscode/*` be added to the gitignore?

PR #896, opened 2025-11-14
refactor MX1 to manually calculate one column, not whole product

StevenClontz:
Similar to how we started asking for one column in MX2, I think we should ask for a random column of the product in MX1. Justification: it reinforces the meaning of the product as a standard matrix for a composition, rather than an arbitrary algorithm.


StevenClontz:
(despite the X, the checks are passing and I think this is ready for review)

Issue #898, opened 2025-11-20
Ensure images and interactives are accessible


siwelwerd:
I believe currently very few of our images have descriptions which will produce alt text, which is necessary to make our books screen reader accessible.  We should make sure every image has a `<shortdescription>`, and probably most of them would benefit from a `<description>` as well.  See [Subsection 4.40.3: Advice for Writing Image Descriptions](https://pretextbook.org/doc/guide/html/topic-accessibility.html#subsec-img-desc-advice) for a longer explanation.

Note that `<shortdescription>`s should be limited to 125 characters and must be plain text.  A `<description>` could use things like `<m>` tags.

Edit: All `<interactive>`s should have a `<description>` as well.  I updated the Style guide to indicate Doenet as the preferred interactive tool (as pending [a bug fix](https://github.com/Doenet/DoenetML/issues/731) they should be keyboard-navigable), or if a 3D interactive is required a Sageplot with a detailed `<description>`.

siwelwerd:
@StevenClontz Does Doenet have any accessibility features at all?  I can't find anything on their website describing how a visusally impaired user might interact with them.  Compared to say https://www.desmos.com/accessibility .

siwelwerd:
Okay, precalculus should be done pending all of the above PRs being merged.

Looks like Calculus has 107 images (compared to 352 for Precal!), so that may go quicker.

siwelwerd:
The merge of #922 completes this.

PR #906, opened 2026-01-12
Image descriptions for Linear Algebra

siwelwerd:
Adding in image descriptions for all of the images in the Linear Algebra book.  Addresses #898 .


siwelwerd:
Thanks for the careful review.  Adding in `<p>` tags seemed to fix things.

PR #911, opened 2026-01-15
Convert Precalculus images to Prefigure

siwelwerd:
Addresses #898 .  This converts the appendix that is the library of graphs into all Prefigure.  This should give templates for converting all of the other graphs in both Precalculus and Calculus.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 pretext[prefigure] == 2.36.0
+prefig >= 0.5.6
 CodeChat-Server == 0.2.25
 pelican == 4.11.0
 markdown == 3.7.0
--- a/source/calculus/source/03-AD/07.ptx
+++ b/source/calculus/source/03-AD/07.ptx
@@ -275,7 +275,7 @@ Note that some of the graphs below are definitely hard... might want to tell thi
                 </task>
                 <task><m>f(x)=\frac{x^2-4}{x^2-9}</m>
                 <answer>
-                Note that there is a horizontal asymptote at <m>y=1</m>.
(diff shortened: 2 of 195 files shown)


siwelwerd:
Ah okay, I think we get the keyboard controls whenever we use `<annotation>`s.  Which should probably be always. But in the `<sidebyside>` it seems to only put them once at the bottom.

siwelwerd:
Hmmm, now it is complaining about `ln` (but succeeding?)

```
error: Unknown function in evaluation: ln
error: Unsafe function definition: ln(x)/ln(2)
error: Error in definition: Unsafe function definition: ln(x)/ln(2)
error: Error retrieving function in graph: Unrecognized name: f
```

[PreFigure documentation](https://prefigure.org/docs/sec-math.html#subsec-math-4) 
 say `ln` is okay

source/precalculus/source/meta/prefigure/graph-logarithm.xml
```diff
@@ -0,0 +1,16 @@
+<diagram dimensions="(300,300)" margins="5" xmlns="https://prefigure.org">
+  <definition>f(x)=ln(x)/ln(2)</definition>
```
siwelwerd:
```suggestion
  <definition>f(x)=log(x)/log(2)</definition>
```

siwelwerd:
It was indeed related.  We were defaulting to an older version of PreFigure, but with the newer PreTeXt that was assuming a newer version of PreFigure.

I feel like the pretext CLI dependency should be telling us what version of prefig to run instead of specifying it manually though (as I did in the last commit to make it work).

PR #914, opened 2026-01-16
Precalculus EQ accessibility

siwelwerd:
Added `<description>`s to all of the number lines (as I am not sure refactoring them into PreFigure adds anything more than the text description).  There were two non-number line images in this chapter (both in EQ3), but they seemed good candidates for interactives so I made those in Doenet.   Addresses #898

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pretext == 2.24.0
+pretext == 2.36.0
 CodeChat-Server == 0.2.25
 pelican == 4.11.0
 markdown == 3.7.0


jkostiuk:
@siwelwerd I'm reviewing the interactives but have a question: in the files-edited section, I can see that there are "descriptions" for the interactives. Where should I see these in the activity book if at all? 

For instance, does Activity 1.3.1 look right to you? I think the activity book is not rendering your description, but I wanted to check with you first about what is "supposed" to be shown.

PR #925, opened 2026-02-10
Update readme

siwelwerd:
There are a few stale things in the Readme that need updating.


README.md
siwelwerd:
```suggestion
```

I believe this is not necessary if using the devcontainer through Docker

siwelwerd:
```suggestion
```

I don't think this is needed for any user not named Steven, so we should remove it from here (and document elsewhere, like in the Wiki, if needed)

jkostiuk (approved):

PR #926, opened 2026-02-10
Tweak language around vertical line test

### LINKED ISSUE #857: Section 2.1 (FN) Vertical Line Test Definitioon
Definition 2.1.17 feels a little awkward to me. I would suggest something like: The **vertical line test** is a method to determine whether a relation is a function by checking whether vertical lines ever intersect the graph in more than one location.

There might be a better way to say it, but saying "a relation on a graph..." just feels a bit weird to me.

siwelwerd:
Closes #857 . I also changed this to a `<fact>` rather than a `<definition>`.

### DIFF
--- a/source/precalculus/source/02-FN/01.ptx
+++ b/source/precalculus/source/02-FN/01.ptx
@@ -570,14 +570,14 @@
     </p>
     </remark>
     
-    <definition xml:id="def-vertical-line-test">
+    <fact xml:id="def-vertical-line-test">
       <statement>
-        <p>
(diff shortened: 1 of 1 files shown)


tdegeorge:
Do we want to say something about how the vertical line test is really just a **visual method** in determining whether a relation is a function? There are other ways of determining a function, but the vertical line test is a nice way to visually see whether it is or is not a function.

siwelwerd:
Well, now that we have Prefigure diagrams, it can produce tactile graphs in a braille version, so it isn't necessarily visual!  😃

PR #930, opened 2026-02-15
Updated Activities 6.5.14 and 6.5.20.

jbunn3:
I found a typo in the latter activity for the 3414.14 number. Then I realized that the A(h) needed a pi removed (it's a square not a disk). Then I fixed the former activity which was missing a `<p> `tag.


siwelwerd (approved):
Looks good, thanks Jared!

PR #933, opened 2026-02-18
Update EV6 class activities to clarify spanning R^n and spanning subspace

StevenClontz:
This is basically the explanation I came up with on the fly today; will try it in my 11am class in a few minutes...

### DIFF
--- a/source/linear-algebra/source/02-EV/06.ptx
+++ b/source/linear-algebra/source/02-EV/06.ptx
@@ -82,7 +82,7 @@ Which feature of <md>\RREF\left[\begin{array}{cccc}
 0&amp;1&amp; 2&amp;0\\
 0&amp;0&amp; 0&amp;1\\
 0&amp;0&amp; 0&amp;0
-\end{array}\right]</md> that shows that <m>W</m>'s spanning set
+\end{array}\right]</md> shows that <m>W</m>'s spanning set
 is linearly dependent?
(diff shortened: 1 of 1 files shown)


siwelwerd:
>will try it in my 11am class in a few minutes...

Perhaps we should await your reflection on its implementation before reviewing then.

source/linear-algebra/source/02-EV/06.ptx
```diff
@@ -68,33 +69,61 @@ the planar subspace.
...
+3&amp;0&amp;-3&amp;5\\
+0&amp;1&amp;2&amp;-1\\
+1&amp;-1&amp;-3&amp;0
+\end{array}\right]=
+\left[\begin{array}{cccc}
+1&amp;0&amp;-1&amp;0\\
+0&amp;1&amp; 2&amp;0\\
+0&amp;0&amp; 0&amp;1\\
+0&amp;0&amp; 0&amp;0
+\end{array}\right]</md> that shows that <m>W</m>'s spanning set
```
jkostiuk:
```suggestion
\end{array}\right]</md> shows that <m>W</m>'s spanning set
```

jkostiuk: found a typo, but otherwise looks good to me!

jkostiuk (changes requested):

PR #936, opened 2026-02-20
Align EV7 activities with EV6

StevenClontz:
IIn class today I found it somewhat clumsy that we don't directly leverage our techniques from EV6 to find a basis in EV7. So I'm going to try this adjustment in my second class.


StevenClontz:
I thought this went well, though I had a bit of deja vu and thought maybe we *used* to do it this way. Let me check.

jkostiuk:
i like it, especially if went well. I think my main desire was for us to have a complex enough example for students to convince themselves the algorithm works--which I think this does!

PR #940, opened 2026-02-25
add more statements to EV1, EV2, EV4, EV5

StevenClontz:
I'm adding lots of various ways to write the statement for tasks X.1 in our EV outcomes, to emphasize there's not one precise way to communciate the answer, while emphasizing that one needs to communicate a full thought.

### DIFF
--- a/source/linear-algebra/exercises/outcomes/EV/EV5/template.xml
+++ b/source/linear-algebra/exercises/outcomes/EV/EV5/template.xml
@@ -16,7 +16,7 @@
                 </p>
             </content>
             <outtro>
-                <p>
+                <p>Here's an example of one such statement:
                 <!-- {{#statementA}} -->
                     <q>The vector equation <m>{{veceqleft}}=\vec w</m> has a unique solution for any <m>\vec w\in\mathbb R^4</m>.</q>
                 <!-- {{/statementA}} -->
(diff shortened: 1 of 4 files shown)


source/linear-algebra/exercises/outcomes/EV/EV5/template.xml
```diff
@@ -17,7 +17,24 @@
...
                 <p>
-                    <q>The vector equation <m>(...)=\vec w</m> has exactly one solution for every <m>\vec w\in\mathbb R^4</m>.</q>
+                <!-- {{#statementA}} -->
+                    <q>The vector equation <m>{{veceqleft}}=\vec w</m> has a unique solution for any <m>\vec w\in\mathbb R^4</m>.</q>
+                <!-- {{/statementA}} -->
+                <!-- {{#statementB}} -->
+                    <q>For any <m>\vec w\in\mathbb R^4</m>, the following vector equation has exactly one solution: <m>{{veceqleft}}=\vec w</m>.</q>
+                <!-- {{/statementB}} -->
+                <!-- {{#statementC}} -->
+                    <q>One and only one solution exists for the equation <m>{{veceqleft}}=\vec w</m> where <m>\vec w</m> is any <m>\mathbb R^4</m> vector.</q>
```
siwelwerd:
```suggestion
                    <q>One and only one solution exists for the equation <m>{{veceqleft}}=\vec w</m> for every vector <m>\vec w</m> in <m>\mathbb R^4</m>.</q>
```

siwelwerd:
Seems like we have multiple tasks and are giving different statements on different tasks, so I think that is probably enough variety as is.  Maybe to Jordan's suggestion just add phrasing like "One such statement is ..."

PR #977, opened 2026-04-24
Fix overfull hboxes in print version

### LINKED ISSUE #962: Fix overfull hboxes
The HTML version allows for horizontally scrollable elements.  Printed pages do not, so sometimes long equations run off the page.

siwelwerd:
Closes #962

### DIFF
--- a/scripts/quickbuild_changes.py
+++ b/scripts/quickbuild_changes.py
@@ -30,7 +30,8 @@ def main():
         xml_ids = []
         for f in changed_files:
             if Path("source", b) in f.parents and f.suffix == ".ptx":
-                root = etree.parse(f).getroot()
+                parser=etree.XMLParser(recover=True)
+                root = etree.parse(f,parser=parser).getroot()
                 if root.tag in ["section", "chapter", "preface", "appendix", "frontmatter"]:
                     xml_id = root.get(r"{http://www.w3.org/XML/1998/namespace}id")
                     xml_ids.append({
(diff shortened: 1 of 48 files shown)


source/linear-algebra/source/01-LE/01.ptx
siwelwerd, line 520: @StevenClontz This is compiling fine by hand, but throwing an error in the preview build.  It complains that there are two images with the same xml:id, but our publication files are all set up to include either the html component (default), or the print component.

Should these images get distinct xml:ids?  Or is there a better way to do this.

scripts/quickbuild_changes.py
```diff
@@ -30,7 +30,8 @@ def main():
         xml_ids = []
         for f in changed_files:
             if Path("source", b) in f.parents and f.suffix == ".ptx":
-                root = etree.parse(f).getroot()
+                parser=etree.XMLParser(recover=True)
+                root = etree.parse(f,parser=parser).getroot()
```
siwelwerd, line 34: @StevenClontz This fixes the error being thrown by duplicate xmlids.  The error was because the xml was being parsed strictly here before pretext was involved.  Please review for any unintended consequences of this change

siwelwerd:
The LaTeX build still reports a few overfull hboxes, but I have looked at them and I think they are okay visually.  So I believe this is ready to go.

jkostiuk (approved):

PR #978, opened 2026-04-26
Fix malformed tags

siwelwerd:
The genius who wrote #974 introduced a bunch of malformed tags, breaking the builds.  This fixes them.


jkostiuk:
lol, sorry for not catching. i thought it was the same kind of "safe" fail as last time.

jkostiuk (approved):

PR #983, opened 2026-04-28
Make page references parenthetical

siwelwerd:
The default way of displaying references in PDF is

image

This changes the behavior to have a parenthetical instead

image

HTML output is unaffected since it does not use this xsl file.

### DIFF
--- a/xsl/print.xsl
+++ b/xsl/print.xsl
@@ -49,4 +49,28 @@
     <xsl:text>fonttitle=\bfseries,</xsl:text>
   </xsl:template>
 
+  <!--Tweak display of references-->
+  <xsl:template match="xref" mode="latex-page-number">
+      <xsl:param name="target"/>
+
+      <xsl:choose>
+          <!-- looks bad when bibliographic number gets wrapped in [] -->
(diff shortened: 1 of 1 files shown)


xsl/print.xsl
```diff
@@ -49,4 +49,28 @@
...
+          <!-- and the number should suffice on its own               -->
+          <xsl:when test="$target/self::biblio"/>
+          <!-- and trailing a () for an equation number is overkill -->
+          <xsl:when test="$target/self::mrow|$target/self::md"/>
+          <!-- and it is really bad for an xref inside a title -->
+          <xsl:when test="ancestor::title"/>
+          <!-- off by default electronic PDF, -->
+          <!-- or on by default for print PDF -->
+          <xsl:when test="not($b-pageref)"/>
+          <!-- OK, requested and helps, let's add it -->
```
siwelwerd, line 67: I'm not sure this is strictly needed, it is copied from the PreTeXt xsl.  I think it is more future-proof to include it though so we are not inadvertently overriding this behavior.

jkostiuk (approved):

PR #994, opened 2026-05-08
Alter presentation of sample exercises

### LINKED ISSUE #993: Tweak title/display of sample problems in print version
Here is the current formatting:
Image

The codes like 'LE3' aren't used anywhere else in the print version so they should be removed in favor of a reference to the corresponding section. I also don't love the numbering "Example B.0.3" but that might be harder to change.

siwelwerd:
Closes #993. This changes the sample exercises to use `<exercise>` instead of `<example>`, and then customizes the typesetting of `<exercise>` in the print version.  It also removes the last references to section codes like "EV3" from the print version, sticking with chapter/section numbers.  HTML version retains the use of these codes

### DIFF
--- a/xsl/print.xsl
+++ b/xsl/print.xsl
@@ -1,8 +1,4 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<!--Copied from entities.ent, used to redefine exercise-->
-<!DOCTYPE xsl:stylesheet [
-    <!ENTITY INLINE-EXERCISE-FILTER "parent::article|parent::paragraphs|parent::chapter|parent::section|parent::subsection|parent::subsubsection|parent::handout">
-]>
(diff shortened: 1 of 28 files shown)


xsl/print.xsl
```diff
@@ -1,4 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" ?>
+<!--Copied from entities.ent, used to redefine exercise-->
+<!DOCTYPE xsl:stylesheet [
+    <!ENTITY INLINE-EXERCISE-FILTER "parent::article|parent::paragraphs|parent::chapter|parent::section|parent::subsection|parent::subsubsection|parent::handout">
+]>
```
siwelwerd:
```suggestion
```

This is vestigial from a previous, clunkier implementation of the same thing. The cleaner implementation below is to just renew the definition of the tcolorbox at the end of the preamble.

jkostiuk (approved):

PR #1004, opened 2026-05-11
Use words

siwelwerd:
Relational operators should not be used outside of properly formatted relations.

### DIFF
--- a/source/linear-algebra/source/02-EV/02.ptx
+++ b/source/linear-algebra/source/02-EV/02.ptx
@@ -162,7 +162,7 @@ since <m>\IR^1=\setBuilder{cx}{c\in\IR}</m>.
   At least <m>m</m> vectors are required to span <m>\IR^m</m>.
         </p>
         <figure>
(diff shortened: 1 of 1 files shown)


source/linear-algebra/source/02-EV/02.ptx
```diff
@@ -162,7 +162,7 @@ since <m>\IR^1=\setBuilder{cx}{c\in\IR}</m>.
   At least <m>m</m> vectors are required to span <m>\IR^m</m>.
         </p>
         <figure>
-            <caption>Failed attempts to span <m>\IR^m</m> by <m>&lt;m</m> vectors</caption>
+            <caption>Failed attempts to span <m>\IR^m</m> by fewer than vectors</caption>
```
jkostiuk:
```suggestion
            <caption>Failed attempts to span <m>\IR^m</m> by fewer than <m>m</m> vectors</caption>
```

jkostiuk (changes requested):

jkostiuk (approved):

PR #1036, opened 2026-05-14
Modular Publication Files

StevenClontz:


publication/includes/html-linear-algebra.ptx
```diff
@@ -0,0 +1,34 @@
...
+        exercise-worksheet="no"
+        exercise-readingquestion="no"
+    />
+    
+    <css theme="denver" palette="leaves"/>
+    <!-- Search can be default or none, or you can use a Google-cx number to use google's search
+    feature (but then variant should be set to none to avoid conflict) -->
+    <search variant="default" />
+    <!-- Set the base URL of where the online version is hosted to have links in other formats -->
+    <baseurl href="https://library.tbil.org/2026/linear-algebra/"/>
```
siwelwerd, line 33: It does not seem to work with `<baseurl>` in the included file, I could only get the right url on QR codes if I put this directly in the publication file.

siwelwerd (changes requested):
See above about `<baseurl>`.  I think I might work around this to get the manuscript to the publisher by the deadline, and suggest this get tabled until the CLI is improved to handle these better.

publication/includes/html-linear-algebra.ptx
StevenClontz: baseurl being included *shouldn't* be a problem because that's all core PreTeXt, and I think it's just the CLI that has a bug

siwelwerd: The CLI is grabbing the baseurl [here](https://github.com/PreTeXtBook/pretext-cli/blob/bb37529ad2274bfd075b2ab50bb859d1942aee54/pretext/project/__init__.py#L76-L83). I confess that I didn't debug enough to verify that this is actually causing the observed problem, but I think that at least disproves your statement that baseurl is all core PreTeXt.

Issue #989, opened 2026-04-30
Interactive QR code URLs resolve to wrong place


siwelwerd:
For example, [Figure 1](https://tbil.org/preview/linear-algebra/LE1.html#figure-le1-add-vectors) in the print version is accompanied by a QR code pointing to the (dead) URL htps://tbil.org/LE1-interactive-add-vectors.html .

siwelwerd:
One fix would be to change the baseurl in the publication file, but then we would have to have different publication files for different books.  Maybe I will just assign this to @StevenClontz to think about.

PR #1039, opened 2026-05-27
Hacky fix for #989

siwelwerd:
This fixes the `<baseurl>` in the publication file that is used for the printed Linear Algebra version.  However, it is a bit hacky as if we want to publish another print book we would have to duplicate this publication file and change just the baseurl.  A better way to do this will be in #1036, but the CLI doesn't seem to handle the `<xi:include>`s correctly right now.

### DIFF
--- a/publication/publication-619.ptx
+++ b/publication/publication-619.ptx
@@ -106,7 +106,7 @@
     javascript, etc).  -->
     <calculator model="none" activecode="none" />
     <!-- Set the base URL of where the online version is hosted to have links in other formats -->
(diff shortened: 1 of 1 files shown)


StevenClontz:
Sorry this got stymied. So the CLI only doesn't xi:include when checking for asset directories, so maybe we can still pull this off. Will get something merged in (or queued up for you) now

StevenClontz (approved):
Actually yeah we should just do this immediately and I'll refactor from here.

PR #1046, opened 2026-08-06
Fix inadvertently commented out outcome (TR4)

siwelwerd:
Apparently the TR4 exercises were not showing up because they are commented out of the bank, but it looks like there are exercises there so this will include them upon rebuild.

### DIFF
--- a/source/precalculus/exercises/bank.xml
+++ b/source/precalculus/exercises/bank.xml
@@ -348,7 +348,6 @@
             </description>
         </outcome>
         
-        <!-- 
         <outcome>
             <title>Special Right Triangles</title>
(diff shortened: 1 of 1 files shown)


jkostiuk (approved):

source/precalculus/exercises/bank.xml
```diff
@@ -348,7 +348,6 @@
             </description>
         </outcome>
         
-        <!-- 
         <outcome>
             <title>Special Right Triangles</title>
```
StevenClontz:
```suggestion
            <title>Special  Right Triangles</title>
```
No op to trigger rebuild

